PR #5, opened 2022-08-10
Solution

Igor-Cegelnyk:


Igor-Cegelnyk:
I can't attach screenshots, raise erorr. I  tried attached file with format jpg and png, but anyway was error "We don't support that file type!"!

Abnormaltype (approved):
Well done!
Try to use only ctrl + c, ctrl + v

PR #55, opened 2022-11-07
Develop

Aleksandro777:
image
image
image
image
image
image
image


shurushku (approved):
Looks ok

Nattalli:
oh, you did git pull, not git rebase?

taxi/urls.py
```diff
@@ -13,12 +13,18 @@
     ManufacturerCreateView,
     ManufacturerUpdateView,
     ManufacturerDeleteView,
+    DriverCreateView,
+    DriverUpdateView,
+    DriverDeleteView, assign_driver_to_car, delete_driver_from_car,
```
Nattalli, line 18: bad code style, make all views from the new line

Nattalli (changes requested):
Good job, left one comment. Also, it looks like you did a git pool, not a git rebase, so other people's commits got pulled up :(

taxi/urls.py
Aleksandro777: Fixed

Aleksandro777:
> 
I did everything as usual. I checked the history in the terminal, maybe I accidentally made a git pull, but no. How do you see that there is a problem? Thank you.

Nattalli (approved):
Of course, this is not a mistake, but it is strange that someone else's commits from the main branch have also reached you. Whether it should be corrected depends on the situation. In this case, it can be left as is. In a real project, either create a new branch and pull possible changes, or make cherry pick?

PR #73, opened 2022-11-08
Solution

LiudmylaKulinchenko:


templates/taxi/driver_update.html
```diff
@@ -0,0 +1,15 @@
...
+{% block content %}
+  <h1>
+    Update drivers license number
+  </h1>
+
+  <form action="" method="post">
+    {% csrf_token %}
+    {{ form|crispy }}
+
+    <input type="submit" value="Submit" class="btn btn-primary">
```
fsocie7y, line 13: i think it's cool add "cancel" button also

fsocie7y: on car creation page the same)

fsocie7y: but it's not a mistake, so i can approve)

fsocie7y (approved):

taxi/urls.py
```diff
@@ -46,6 +51,27 @@
     path(
         "drivers/<int:pk>/", DriverDetailView.as_view(), name="driver-detail"
     ),
+    path("drivers/create/", DriverCreateView.as_view(), name="driver-create"),
+    path(
+        "drivers/<int:pk>/update",
```
KirillDrago, line 56: you forgot to add "/" at the end
Everything else looks good as for me)

KirillDrago (changes requested):

taxi/urls.py
LiudmylaKulinchenko: I made same mistake 😂

KirillDrago (approved):

PR #91, opened 2022-11-09
Solution

Ilyakson:


taxi/views.py
```diff
@@ -87,3 +90,30 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+
+
+class DriverLicenseUpdateView(LoginRequiredMixin, generic.UpdateView):
+    model = Driver
+    form_class = DriverLicenseUpdateForm
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+@login_required
+def link_add(request, pk):
```
mltdll, line 112: It looks like you've changed the implementation but forgot the function's name. This function combines two purposes now.

mltdll (changes requested):

mltdll:
By the way, it looks like I've seen your solution before. I think it would be better if you committed your own implementation before taking it from the following task.

mltdll (approved):

TPolina (approved):

PR #105, opened 2022-12-09
Solution

GeorgePavlej:
assigh
create car
create driver
delete driver
delete
update license number button
update license


OleksandrPozden (approved):
Well done!

PR #132, opened 2023-02-17
Solution

Ihor-Ivashov:

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,9 +1,12 @@
+from django.contrib.auth import get_user_model
 from django.contrib.auth.decorators import login_required
-from django.shortcuts import render
-from django.urls import reverse_lazy
+from django.http import HttpResponseRedirect
+from django.shortcuts import render, get_object_or_404
+from django.urls import reverse_lazy, reverse
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
 
+from .forms import DriverForm, DriverLicenseUpdateForm, CarForm
 from .models import Driver, Car, Manufacturer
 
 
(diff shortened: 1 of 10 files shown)


taxi/views.py
```diff
@@ -87,3 +90,29 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    form_class = DriverLicenseUpdateForm
+    template_name = "taxi/driver_license_update_form.html"
+
+
+class DriverDeleteView(LoginRequiredMixin, generic.DeleteView):
+    model = get_user_model()
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+def cars_add_delete_driver(request, pk):
```
ArsenPidhoretskyi, line 111: Rewrite to class)

taxi/views.py
```diff
@@ -64,14 +67,14 @@ class CarDetailView(LoginRequiredMixin, generic.DetailView):
 
 class CarCreateView(LoginRequiredMixin, generic.CreateView):
     model = Car
-    fields = "__all__"
-    success_url = reverse_lazy("taxi:car-list")
+    form_class = CarForm
+    # success_url = reverse_lazy("taxi:car-list")
```
ArsenPidhoretskyi, line 71: Why you comment this?

ArsenPidhoretskyi (changes requested):
Add new line at the end of html files

taxi/views.py
Ihor-Ivashov: Because i want to navigate to detail page after creation

taxi/views.py
Ihor-Ivashov: I don`t know how to do it

Grayder0152 (approved):

PR #157, opened 2023-03-21
1 commit

AndereLion:

### DIFF
--- a/taxi/urls.py
+++ b/taxi/urls.py
@@ -47,7 +47,7 @@ urlpatterns = [
     path("cars/<int:pk>/update/", CarUpdateView.as_view(), name="car-update"),
     path("cars/<int:pk>/delete/", CarDeleteView.as_view(), name="car-delete"),
     path(
-        "cars/<int:pk>/assign-delete-relation/",
+        "cars/<int:pk>/assign/",
         assign_unassign_to_car,
         name="assign_or_unassign_to_car",
     ),
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,5 +1,6 @@
 from django.contrib.auth.decorators import login_required
 from django.contrib.auth.mixins import LoginRequiredMixin
+from django.http import HttpResponseRedirect
 from django.shortcuts import render, redirect, get_object_or_404
 from django.urls import reverse_lazy
 from django.views import generic
@@ -106,12 +107,14 @@ class DriverLicenseDeleteView(LoginRequiredMixin, generic.DeleteView):
     success_url = reverse_lazy("taxi:driver-list")
 
 
+@login_required()
 def assign_unassign_to_car(request, *args, **kwargs):
     car_pk = kwargs["pk"]
-    users_cars = request.user.cars
-    get_object_or_404(Car, id=car_pk)
-    if users_cars.filter(id=car_pk):
-        users_cars.remove(car_pk)
+    car = Car.objects.get(id=kwargs["pk"])
+    driver = Driver.objects.get(id=request.user.id)
+
+    if car in driver.cars.all():
+        driver.cars.remove(car_pk)
     else:
-        users_cars.add(car_pk)
-    return redirect("taxi:car-detail", pk=car_pk)
+        driver.cars.add(car_pk)
+    return HttpResponseRedirect(reverse_lazy("taxi:car-detail", args=[car_pk]))


taxi/views.py
```diff
@@ -87,3 +88,29 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+
+
+class DriverLicenseDeleteView(LoginRequiredMixin, generic.DeleteView):
+    model = Driver
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+def assign_unassign_to_car(request, *args, **kwargs):
+    car_pk = kwargs["pk"]
+    user = request.user.cars
```
SafonovVladimir, line 111: I guess user_cars or smth like that will be better name

SafonovVladimir (approved):
GJ

taxi/urls.py
```diff
@@ -42,7 +46,24 @@
     path("cars/create/", CarCreateView.as_view(), name="car-create"),
     path("cars/<int:pk>/update/", CarUpdateView.as_view(), name="car-update"),
     path("cars/<int:pk>/delete/", CarDeleteView.as_view(), name="car-delete"),
+    path(
+        "cars/<int:pk>/assign-unassign/",
```
wQuelS, line 50: choose a better name for endpoint, since `assign-unassign` is a bit too overloaded

taxi/views.py
```diff
@@ -86,4 +87,30 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    if user.filter(id=car_pk):
+        user.remove(car_pk)
+    else:
+        user.add(kwargs["pk"])
```
wQuelS, line 115: u've extracted this `kwargs["pk"]` to `car_pk` variable, so why don't you use it here ?

taxi/views.py
```diff
@@ -86,4 +87,30 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+
+
+class DriverLicenseDeleteView(LoginRequiredMixin, generic.DeleteView):
+    model = Driver
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+def assign_unassign_to_car(request, *args, **kwargs):
+    car_pk = kwargs["pk"]
+    user = request.user.cars
```
wQuelS, line 111: user?

taxi/views.py
```diff
@@ -86,4 +87,30 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    if user.filter(id=car_pk):
```
wQuelS, line 112: it is better to use `get` since you need only one record

wQuelS (changes requested):

taxi/urls.py
AndereLion: Fixed

taxi/views.py
AndereLion: Fixed =)

taxi/views.py
AndereLion: I guess filter here is better than using try except? But if you insist i will rewrite it. =)

taxi/views.py
AndereLion: Fixed

taxi/urls.py
```diff
@@ -42,7 +46,24 @@
     path("cars/create/", CarCreateView.as_view(), name="car-create"),
     path("cars/<int:pk>/update/", CarUpdateView.as_view(), name="car-update"),
     path("cars/<int:pk>/delete/", CarDeleteView.as_view(), name="car-delete"),
+    path(
+        "cars/<int:pk>/assign-delete-relation/",
```
wQuelS, line 50: i guess it should be `toggle` or just `assign`

taxi/views.py
```diff
@@ -86,4 +87,31 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+
+def assign_unassign_to_car(request, *args, **kwargs):
+    car_pk = kwargs["pk"]
+    users_cars = request.user.cars
+    get_object_or_404(Car, id=car_pk)
+    if users_cars.filter(id=car_pk):
+        users_cars.remove(car_pk)
+    else:
+        users_cars.add(car_pk)
+    return redirect("taxi:car-detail", pk=car_pk)
```
wQuelS, line 117: use HttpResponseRedirect with `reverse_lazy()` and provide pk in `args=`. It would be much more better

taxi/views.py
```diff
@@ -86,4 +87,31 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    form_class = DriverLicenseUpdateForm
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+class DriverLicenseDeleteView(LoginRequiredMixin, generic.DeleteView):
+    model = Driver
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+def assign_unassign_to_car(request, *args, **kwargs):
```
wQuelS, line 109: I guess this func should be able only for logged in users. What do you think?)

taxi/views.py
```diff
@@ -86,4 +87,31 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    car_pk = kwargs["pk"]
+    users_cars = request.user.cars
+    get_object_or_404(Car, id=car_pk)
```
wQuelS, line 112: just extract `car` and `driver` into variables with this method. and then use something like that `if driver in car.drivers.all()`.
Would be much more easier to read and you will get rid of redundant lines

wQuelS (changes requested):

taxi/views.py
AndereLion: Fixed all

wQuelS (approved):
GJ

PR #161, opened 2023-03-23
Solution

MaybeTI:
image
image
image
image
image


SergiiMachulin (changes requested):
You've done Assign/Delete buttons option on the drivers detail page, but this must be done on the car detail page according to p.6 README.md. Also you need to check outlook of desired page according to the provided picture in README.md (Username, Full name, License number)

PR #162, opened 2023-03-24
solution

DavidGlusch:

### DIFF
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -1,0 +1,45 @@
+from django import forms
+from django.contrib.auth import get_user_model
+from django.core.exceptions import ValidationError
+
+from taxi.models import Driver, Car
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    LENGTH = 8
+
+    class Meta:
+        model = Driver
+        fields = ("license_number", )
+
+    def clean_license_number(self) -> str:
+        license_number = self.cleaned_data["license_number"]
+
+        if len(license_number) != DriverLicenseUpdateForm.LENGTH:
+            raise ValidationError(
+                f"License number should be "
+                f"{DriverLicenseUpdateForm.LENGTH} symbols"
+            )
+
+        if not license_number[:3].isupper():
+            raise ValidationError("First 3 characters should be capitalized")
+
+        if not license_number[:3].isalpha():
+            raise ValidationError("First 3 characters should be letters")
+
+        if not license_number[-5:len(license_number):1].isdigit():
+            raise ValidationError("Last 5 characters should be digits")
+
+        return license_number
+
+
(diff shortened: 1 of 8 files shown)


ZabFTFT (approved):
GJ!

YuriiKindrat (changes requested):

taxi/views.py
```diff
@@ -87,3 +88,30 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+
+
+class DriverDeleteView(LoginRequiredMixin, generic.DeleteView):
+    model = Driver
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+def assign_or_delete(request, pk):
+    car = Car.objects.get(id=pk)
+    driver = Driver.objects.get(id=request.user.id)
```
YuriiKindrat, line 112: no need to make query here, u can just use request.user instead

YuriiKindrat (changes requested):

taxi/forms.py
```diff
@@ -0,0 +1,45 @@
...
+        if not license_number[:3].isupper():
+            raise ValidationError("First 3 characters should be capitalized")
+
+        if not license_number[:3].isalpha():
+            raise ValidationError("First 3 characters should be letters")
+
+        if not license_number[-5:len(license_number):1].isdigit():
+            raise ValidationError("Last 5 characters should be digits")
+
+        return license_number
```
wQuelS, line 33: Still no validation for DriverCreation"?

wQuelS (changes requested):

taxi/forms.py
DavidGlusch: as i understand all i had to do was use this form for driver creation view as i did, and i cant change this form name since it fails tests if i do, should i create separate creation form validator?

wQuelS: Just create an additional `DriverCreationForm` which inherits from `UserCreationForm`. And implement validation on creation

wQuelS (changes requested):
Just create an additional DriverCreationForm which inherits from UserCreationForm. And implement validation on creation

taxi/forms.py
```diff
@@ -1,16 +1,45 @@
 from django import forms
 from django.contrib.auth import get_user_model
+from django.contrib.auth.forms import UserCreationForm
 from django.core.exceptions import ValidationError
 
 from taxi.models import Driver, Car
 
 
+class DriverLicenseCreateForm(UserCreationForm):
```
wQuelS, line 9: it is not a License, but DriverCreation form. When Driver creates - what makes him a "driver" actually?
Also, now only this form has validation, what about UpdateLicenseForm?

wQuelS (changes requested):

PR #163, opened 2023-03-24
Solution

ZabFTFT:
crease driver
create driver form
update delete
license number error
delete me from car
assign to car


DmytryTong (approved):
Seems 2b workable)

taxi/forms.py
```diff
@@ -0,0 +1,61 @@
...
+        )
+
+
+class DriverLicenseUpdateForm(forms.ModelForm, LicenseNumberValidationMixin):
+    class Meta:
+        model = Driver
+        fields = ("license_number",)
+
+
+class CarCreateForm(forms.ModelForm):
```
wQuelS, line 52: you will use this form to create and update also. Choose more accurate name, please

taxi/views.py
```diff
@@ -87,3 +102,21 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+class DriverDeleteView(LoginRequiredMixin, generic.DeleteView):
+    model = Driver
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+class DriverLicenseUpdateView(LoginRequiredMixin, generic.UpdateView):
+    model = Driver
+    field = ["license_number"]
+    form_class = DriverLicenseUpdateForm
+    template_name = "taxi/license_form.html"
```
wQuelS, line 122: you could use just `DriverLicenseUpdateForm` without additional template for `license_form` and make everything in driver_form using `object|yesno`

wQuelS (changes requested):

taxi/views.py
ZabFTFT: Thank you for this idea. I will implement this a little bit later ;)

PR #174, opened 2023-03-26
'Solution'

Morty67:


Morty67:
about_driver

Morty67:
update car

Morty67:
update_license

Morty67:
delete_driver

Morty67:
create new driver

Morty67:
Screenshot_1

taxi/forms.py
```diff
@@ -0,0 +1,57 @@
...
+    DRIVER_LICENSE_LENGTH = 8
+
+    class Meta:
+        model = Driver
+        fields = ("license_number",)
+
+    def clean_license_number(self):
+        license_number = self.cleaned_data["license_number"]
+        if len(license_number) != DriverLicenseUpdateForm.\
+                DRIVER_LICENSE_LENGTH:
```
YuriiKindrat, line 29: it seems to me, it is better just to write an 8 here, to avoid line breaking, or use much smaller name for constant

taxi/views.py
```diff
@@ -28,6 +37,20 @@ def index(request):
...
+    car = Car.objects.get(pk=pk)
+    car.drivers.add(request.user)
+    return redirect("taxi:car-detail", pk=car.pk)
+
+
+@login_required
+def delete_from_car(request: HttpRequest, pk: int) -> HttpResponse:
+    car = Car.objects.get(id=pk)
+    car.drivers.remove(request.user)
+    return redirect("taxi:car-detail", pk=car.pk)
```
YuriiKindrat, line 51: try to make only 1 view for this

taxi/views.py
```diff
@@ -87,3 +110,21 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+class DriverDeleteView(LoginRequiredMixin, generic.DeleteView):
+    model = Driver
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+class DriverUpdateView(LoginRequiredMixin, generic.UpdateView):
+    model = Driver
+    form_class = DriverLicenseUpdateForm
+    success_url = reverse_lazy("taxi:driver-list")
+    template_name = "taxi/driver_update.html"
```
YuriiKindrat, line 130: won't it work without template name ?

YuriiKindrat (changes requested):

taxi/views.py
```diff
@@ -28,6 +37,18 @@ def index(request):
     return render(request, "taxi/index.html", context=context)
 
 
+@login_required
+def toggle_assign_to_car(request, pk):
+    driver = Driver.objects.get(id=request.user.id)
```
YuriiKindrat, line 42: no need to make query here, u can use request.user instead

YuriiKindrat (changes requested):

PR #181, opened 2023-03-27
Forms Solution

Maytreyaya:
licensde
user_page
assighn
delete
dr_list
driver_choose

### DIFF
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -6,13 +6,11 @@ from django.core.exceptions import ValidationError
 from taxi.models import Driver, Car
 
 
-class DriverLicenseUpdateForm(forms.ModelForm):
-    class Meta:
-        model = Driver
-        fields = ("license_number",)
-
+class CleanLicenseMixin:
     def clean_license_number(self):
+
         num = self.cleaned_data["license_number"]
+
         if (len(num) != 8
                 or not num[:3].isupper()
                 or not num[3:].isdigit()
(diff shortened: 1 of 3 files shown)


ada-krav (approved):

taxi/forms.py
```diff
@@ -0,0 +1,49 @@
...
+                or not num[3:].isdigit()
+                or not num[:3].isalpha()):
+            raise ValidationError(
+                "License Number must consist only of 8 chars,"
+                " first 3 chars must be in uppercase,"
+                " last 3 must be digits")
+        return num
+
+
+class DriverForm(UserCreationForm, DriverLicenseUpdateForm):
```
wQuelS, line 27: It is weird that the more general `DriverForm` inherits from the more specific `DriverLicenseUpdateForm`. Consider to use `composition` design pattern here. One way to do this is by creating a mixin that contains the `clean_license_number`

taxi/urls.py
```diff
@@ -39,13 +43,43 @@
...
+
+    path("drivers/<int:pk>/delete/",
+         DriverDeleteView.as_view(),
+         name="driver-delete"),
+
+    path("drivers/<int:pk>/update/",
+         LicenseUpdateView.as_view(),
+         name="driver-update"),
+
+    path("cars/<int:pk>/assign-or-delete-to-car/",
```
wQuelS, line 79: Naming here is not good.  Make it more concise

taxi/views.py
```diff
@@ -87,3 +90,35 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
 class DriverDetailView(LoginRequiredMixin, generic.DetailView):
     model = Driver
     queryset = Driver.objects.all().prefetch_related("cars__manufacturer")
+
+
+class DriverCreateView(LoginRequiredMixin, generic.CreateView):
+    model = Driver
+    # fields = "__all__"
```
wQuelS, line 97: remove redundant comment

taxi/views.py
```diff
@@ -87,3 +90,35 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    form_class = DriverForm
+    template_name = "taxi/driver_form.html"
```
wQuelS, line 99: it is redundant. re-read checklist and fix everywhere

wQuelS (changes requested):

YuriiKindrat (approved):

PR #186, opened 2023-03-27
solution

inrock1:


ProdyRodion (approved):
Good job!

taxi/views.py
```diff
@@ -81,9 +83,38 @@ class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
...
+class DriverCreateView(LoginRequiredMixin, generic.CreateView):
+    model = Driver
+    form_class = DriverCreationForm
+
+
+class DriverDeleteView(LoginRequiredMixin, generic.DeleteView):
+    model = Driver
+    fields = "__all__"
+    success_url = reverse_lazy("taxi:driver-list")
+    template_name = "taxi/driver_confirm_delete.html"
```
YuriiKindrat, line 103: won't it work without template_name ?

YuriiKindrat (changes requested):

taxi/views.py
inrock1: it will work.
but for first project it is better for me to keep a similar template name with comment.
have I gotta remove it?

inrock1: maybe, it won't work.
because it not similar with `myapp/<model_name>_form.html` .
comment deleted.

taxi/views.py
```diff
@@ -1,8 +1,10 @@
 from django.contrib.auth.decorators import login_required
+from django.http import HttpResponseRedirect
 from django.shortcuts import render
 from django.urls import reverse_lazy
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
+from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm
 
 from .models import Driver, Car, Manufacturer
```
wQuelS, line 9: make it according to PEP8

taxi/views.py
```diff
@@ -81,9 +83,38 @@ class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
...
+class DriverCreateView(LoginRequiredMixin, generic.CreateView):
+    model = Driver
+    form_class = DriverCreationForm
+
+
+class DriverDeleteView(LoginRequiredMixin, generic.DeleteView):
+    model = Driver
+    fields = "__all__"
+    success_url = reverse_lazy("taxi:driver-list")
+    template_name = "taxi/driver_confirm_delete.html"
```
wQuelS, line 103: it will work without providing a template name

wQuelS (changes requested):

taxi/views.py
inrock1: ok, then can I leave it with a comment?
because this name is not obvious to me.

taxi/views.py
inrock1: sure. fixed

taxi/views.py
```diff
@@ -81,9 +86,38 @@ class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
...
```
wQuelS, line 106: so is working without providing template_name?

wQuelS (changes requested):

wQuelS (approved):
GJ

PR #205, opened 2023-03-28
Solution

tytomyr:
image

image

image

### DIFF
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -1,0 +1,45 @@
+from django import forms
+from django.contrib.auth import get_user_model
+from django.core.exceptions import ValidationError
+from taxi.models import Driver, Car
+
+
+class DriverUpdateForm(forms.ModelForm):
+    class Meta:
+        model = Driver
+        fields = "__all__"
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    class Meta:
+        model = Driver
+        fields = ("license_number",)
+
+    def clean_license_number(self):
+        license_number = self.cleaned_data["license_number"]
+        first = license_number[:3]
+        last = license_number[-5:]
+        length = len(license_number)
+
+        def contains_number(string):
+            return any(char.isdigit() for char in string)
+
+        if (length != 8
+                or first.upper() != first
+                or not last.isdigit()
+                or contains_number(first)):
+            raise ValidationError("Ensure that value is valid")
+
+        return license_number
+
+
+class CarUpdateForm(forms.ModelForm):
+    drivers = forms.ModelMultipleChoiceField(
+        queryset=get_user_model().objects.all(),
+        widget=forms.CheckboxSelectMultiple,
+        required=False
+    )
+
+    class Meta:
+        model = Car
+        fields = "__all__"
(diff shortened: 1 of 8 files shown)


taxi/forms.py
```diff
@@ -0,0 +1,45 @@
...
+        fields = ("license_number",)
+
+    def clean_license_number(self):
+        license_number = self.cleaned_data["license_number"]
+        first = license_number[:3]
+        last = license_number[-5:]
+        length = len(license_number)
+
+        def contains_number(string):
+            return any(char.isdigit() for char in string)
```
YuriiKindrat, line 25: try to use isalpha method )

taxi/forms.py
```diff
@@ -0,0 +1,45 @@
...
+
+        if (length != 8
+                or first.upper() != first
```
YuriiKindrat, line 28: u can use isupper method here

taxi/forms.py
```diff
@@ -0,0 +1,45 @@
...
+                or not last.isdigit()
+                or contains_number(first)):
+            raise ValidationError("Ensure that value is valid")
```
YuriiKindrat, line 31: write more descriptive message for error

taxi/views.py
```diff
@@ -64,16 +65,30 @@ class CarDetailView(LoginRequiredMixin, generic.DetailView):
...
+    driver = Driver.objects.get(id=request.user.pk)
+    driver.cars.add(pk)
+    return redirect(reverse_lazy("taxi:car-list"))
+
+
+@login_required
+def car_remove_driver(request, pk):
+    driver = Driver.objects.get(id=request.user.pk)
+    driver.cars.remove(pk)
+    return redirect(reverse_lazy("taxi:car-list"))
```
YuriiKindrat, line 89: try to make this using one view

taxi/forms.py
```diff
@@ -0,0 +1,45 @@
+from django import forms
+from django.contrib.auth import get_user_model
+from django.core.exceptions import ValidationError
+from taxi.models import Driver, Car
+
+
+class DriverUpdateForm(forms.ModelForm):
+    class Meta:
+        model = Driver
+        fields = "__all__"
```
YuriiKindrat, line 10: why do u use this form for creating ? check ur checklist

YuriiKindrat (changes requested):

taxi/views.py
```diff
@@ -64,16 +65,26 @@ class CarDetailView(LoginRequiredMixin, generic.DetailView):
...
 
 
+@login_required
+def car_add_or_remove_driver(request, pk):
+    driver = Driver.objects.get(id=request.user.pk)
+    if driver in Driver.objects.all():
+        driver.cars.remove(pk)
+    else:
+        driver.cars.add(pk)
+    return redirect(reverse_lazy("taxi:car-list"))
```
YuriiKindrat, line 85: is it works as expected ? U are getting driver by id and then check if it exist in list of all drivers, am i right ? 
fix that, also no need to get driver here, u can use reuqest.user instead

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #206, opened 2023-03-28
Solution

Zenionoi:


Zenionoi:
image
image
image
image

taxi/views.py
```diff
@@ -87,3 +89,31 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+
+class DriverDeleteView(LoginRequiredMixin, generic.DeleteView):
+    model = Driver
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+@login_required
+def driver_assign(request, pk):
+    car = Car.objects.get(pk=pk)
+    driver = Driver.objects.get(pk=request.user.pk)
```
YuriiKindrat, line 114: u can use request.user instead

YuriiKindrat (changes requested):

wQuelS (changes requested):
Fix indentation in your html files. It is better to use 2 tabs

PR #212, opened 2023-03-29
Solution

yhwh6:


taxi/forms.py
```diff
@@ -0,0 +1,55 @@
...
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    class Meta:
+        model = Driver
+        fields = ("license_number",)
+
+    def clean_license_number(self) -> str:
+        license_number = self.cleaned_data["license_number"]
+        license_number_validator(license_number)
+        return license_number
```
ArsenPidhoretskyi, line 55: Duplicate code. Maybe create one more class(mixin) and then inherit from it?

taxi/views.py
```diff
@@ -87,3 +92,34 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+class DriverDeleteView(LoginRequiredMixin, generic.DeleteView):
+    model = Driver
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+@login_required
+def delete_car_assignment(request, pk):
```
ArsenPidhoretskyi, line 115: Write as class view

ArsenPidhoretskyi (changes requested):

taxi/views.py
```diff
@@ -87,3 +92,35 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+
+
+class DriverDeleteView(LoginRequiredMixin, generic.DeleteView):
+    model = Driver
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+class CarAssignmentView(LoginRequiredMixin, generic.View):
+    @staticmethod
+    def get(request, pk):
```
ArsenPidhoretskyi, line 116: Do this on post method. Not good to change object on get method

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Good job!

PR #213, opened 2023-03-31
Solution

StreetNik:
Screenshot 2023-03-30 at 22 23 32
Screenshot 2023-03-30 at 22 23 50
Screenshot 2023-03-30 at 22 23 40

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,9 +1,11 @@
 from django.contrib.auth.decorators import login_required
+from django.http import HttpResponseRedirect
 from django.shortcuts import render
-from django.urls import reverse_lazy
+from django.urls import reverse_lazy, reverse
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
 
+from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarCreateForm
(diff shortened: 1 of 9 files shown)


taxi/views.py
```diff
@@ -62,18 +64,39 @@ class CarDetailView(LoginRequiredMixin, generic.DetailView):
     model = Car
 
 
+def car_driver_assign_or_delete(request, pk):
```
ArsenPidhoretskyi, line 67: Please rewrite as class view

taxi/views.py
```diff
@@ -62,18 +64,39 @@ class CarDetailView(LoginRequiredMixin, generic.DetailView):
...
+    if request.method == "POST":
+        driver_id = request.POST["driver_id"]
+        car = Car.objects.get(id=pk)
+        remove = request.POST.get("remove")
```
ArsenPidhoretskyi, line 71: Maybe better just to check if the driver inside car.drivers? Remove redundant variable

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #237, opened 2023-05-20
Solution

bohdan-yatsyna:

### DIFF
--- a/taxi/views.py
+++ b/taxi/views.py
@@ -1,10 +1,15 @@
 from django.contrib.auth.decorators import login_required
-from django.shortcuts import render
+from django.contrib.auth.mixins import LoginRequiredMixin
+from django.shortcuts import render, redirect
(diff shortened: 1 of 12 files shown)


taxi/views.py
```diff
@@ -79,6 +84,19 @@ class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
     success_url = reverse_lazy("taxi:car-list")
 
 
+@login_required
+def car_add_or_delete_driver(request, pk):
```
ArsenPidhoretskyi, line 88: Write as class view

ArsenPidhoretskyi (changes requested):

taxi/views.py
bohdan-yatsyna: Thank you for the remark, done.

vasylhnatiuk (approved):

PR #275, opened 2023-07-07
Solution

nailia-mamiedova:


nailia-mamiedova:
image
Drivers List

image
Driver Detail

image
Update License Number

image
Delete Driver

image
In car.drivers

image
Not in car.drivers

PR #289, opened 2023-07-17
'Solution'

ant-komarov:
image
image
image
image
image
image
image
image


taxi/views.py
```diff
@@ -79,11 +79,40 @@ class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
     success_url = reverse_lazy("taxi:car-list")
 
 
+def car_update_driver_view(request, pk):
+    car = Car.objects.get(id=pk)
+    driver = Driver.objects.get(id=request.user.id)
```
YuriiKindrat, line 84: u can use request.user instead, no need to get driver from db

YuriiKindrat (changes requested):

Oleg-s-s (approved):
:+1:

PR #293, opened 2023-07-18
Solution

Kirontiko:
Assigning to car

image

Deleting from car

image

Create new driver button
image

Validating

image

image

Driver update license number button and delete driver button

image

Delete confirmation button

image

Checkbox in the car creation page

image


Oleg-s-s (approved):
:+1:

PR #310, opened 2023-07-20
'Solution'

RedCatHot:
image
image
image
image


taxi/urls.py
```diff
@@ -46,6 +49,22 @@
...
+    path(
+        "drivers/<int:pk>/delete/",
+        DriverDeleteView.as_view(),
+        name="driver-delete"
+    ),
+    path(
+        "cars/<int:pk>/assign_delete/",
+        CarDetailView.as_view(),
+        name="car-assign-delete"
+    ),
```
YuriiKindrat: there is no need to create a new link here

templates/taxi/driver_confirm_delete.html
```diff
@@ -0,0 +1,10 @@
+{% extends "base.html" %}
+
+{% block content %}
+  <h1>Delete driver?</h1>
+  <form action="" method="post">
+    {% csrf_token %}
+
+    <input type="submit" value="Yes" class="btn btn-danger">
+  </form>
+{% endblock %}
```
YuriiKindrat: no blank line 
fix this everywhere

YuriiKindrat (changes requested):

templates/taxi/driver_confirm_delete.html
RedCatHot: Fixed, added blank lines in some html files

taxi/urls.py
RedCatHot: Why not? May you explain please?

Oleg-s-s (approved):
:+1:

PR #334, opened 2023-08-31
Solution

prykhodkom:


prykhodkom:
2023-08-31 19_22_40-Taxi Service
2023-08-31 19_23_08-Taxi Service

taxi/forms.py
```diff
@@ -0,0 +1,52 @@
...
+from django.core.exceptions import ValidationError
+
+from taxi.models import Driver, Car
+
+
+def validate_license(license_number: str):
+    length = (len(license_number) == 8)
+    letters = all(
+        char.isalpha() and char.isupper() for char in license_number[:3]
+    )
```
olena-hul: You can check the whole string instead of checking every char

taxi/forms.py
```diff
@@ -0,0 +1,52 @@
...
+    digits = all(char.isnumeric() for char in license_number[3:])
```
olena-hul: You can check the whole string instead of checking every char

olena-hul (changes requested):

Bohdan-Salamakha (changes requested):
fix tests

Bohdan-Salamakha (approved):

PR #360, opened 2023-09-27
Solution

ochernous:
taxi-service-user-form-and-validation1
taxi-service-user-form-and-validation2
taxi-service-user-form-and-validation3
taxi-service-user-form-and-validation4
taxi-service-user-form-and-validation5
taxi-service-user-form-and-validation6
taxi-service-user-form-and-validation7
taxi-service-user-form-and-validation8
taxi-service-user-form-and-validation9

### DIFF
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -1,0 +1,53 @@
+from django import forms
+from django.contrib.auth import get_user_model
+from django.contrib.auth.forms import UserCreationForm
+from django.core.exceptions import ValidationError
+
+from taxi.models import Driver, Car
+
+
+def clean_license_number(cleaned_data):
+    license_number_length = 8
+    license_number = cleaned_data["license_number"]
+    if len(license_number) != license_number_length:
+        raise ValidationError(
+            "The license number should be "
+            f"{license_number_length} characters long"
+        )
+    if not license_number[:3].isupper() or not license_number[:3].isalpha():
+        raise ValidationError("First 3 symbols should be capital letters")
+    if not license_number[3:].isdigit():
+        raise ValidationError("Last 5 symbols should be numbers")
+    return license_number
+
+
+class DriverCreationForm(UserCreationForm):
+    class Meta(UserCreationForm.Meta):
+        model = Driver
+        fields = UserCreationForm.Meta.fields + (
+            "first_name", "last_name", "license_number",
+        )
+
+    def clean(self):
+        clean_license_number(self.cleaned_data)
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    class Meta:
+        model = Driver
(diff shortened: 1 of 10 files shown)


templates/taxi/car_detail.html
```diff
@@ -2,21 +2,41 @@
...
+  <br>
   <p>Manufacturer: ({{ car.manufacturer.name }}, {{ car.manufacturer.country }})</p>
-  <h1>Drivers</h1>
   <hr>
-  <ul>
+
+  <div class="ml-3">
+    <h4>
+      Drivers
+      {% if user in car.drivers.all %}
```
denys-source: `car.drivers.all` hits db 2 times on this page. it is better to use `with` or add this queryset to the context in the corresponding view

taxi/views.py
```diff
@@ -87,3 +89,37 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
 class DriverDetailView(LoginRequiredMixin, generic.DetailView):
     model = Driver
     queryset = Driver.objects.all().prefetch_related("cars__manufacturer")
```
denys-source: you can skip `all()` here

taxi/views.py
```diff
@@ -87,3 +89,37 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    model = Driver
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+@login_required
+def assign_or_delete(request: HttpRequest, car_id: int) -> HttpResponse:
+    user = request.user
+    car = Car.objects.get(id=car_id)
+
+    if user in car.drivers.all():
```
denys-source: if `user` in `car.drivers.all()` you can just delete it without any additional checks

taxi/forms.py
```diff
@@ -0,0 +1,53 @@
...
+
+class DriverCreationForm(UserCreationForm):
+    class Meta(UserCreationForm.Meta):
+        model = Driver
+        fields = UserCreationForm.Meta.fields + (
+            "first_name", "last_name", "license_number",
+        )
+
+    def clean(self):
+        clean_license_number(self.cleaned_data)
```
denys-source: I think it is better to use mixin with `clean_<field_name>` method or create validator and use it in both forms

Barsh4ec (approved):
I agree with Denys

Serhii-Leonenko (changes requested):
Fix according to your colleague's comments

taxi/forms.py
Sofbog: Agree, it will make code more clear

Sofbog (changes requested):

Serhii-Leonenko (approved):

PR #410, opened 2023-11-20
Solution

Naz-iv:


Naz-iv:
image
image
image
image
image
image

taxi/forms.py
```diff
@@ -0,0 +1,60 @@
...
+        pattern = r"^[A-Z]{3}[0-9]{5}"
+
+        if (len(license_number) != self.LICENSE_REQUIRED_LENGTH
+                or not re.match(pattern=pattern, string=license_number)):
+            raise ValidationError(
+                "License number invalid! Please enter correct license number."
+            )
+
+        return license_number
+
```
Familenko: I do not see the reason to make two identical form
There only diference is fields, but in this case you can inheret from other class
```
class DriverCreationForm(UserCreationForm, DriverLicenseUpdateForm):
    fields = "first_name", "last_name", "license_number",
    model = get_user_model()
```

taxi/forms.py
```diff
@@ -0,0 +1,60 @@
...
+    def clean_license_number(self):
+        license_number = self.cleaned_data["license_number"]
+        pattern = r"^[A-Z]{3}[0-9]{5}"
+
+        if (len(license_number) != self.LICENSE_REQUIRED_LENGTH
+                or not re.match(pattern=pattern, string=license_number)):
+            raise ValidationError(
+                "License number invalid! Please enter correct license number."
+            )
+
```
Familenko, line 35: Maybe it is better to specify every particular error for better user expirience

Familenko (changes requested):
On your screenshot jim.hoper have license number == JIM26531122
So maybe your make screenshot before attempt validation or validation is not working
Or maybe it was automaticaly installed with database, just kill him

taxi/forms.py
Naz-iv: Jim never made it to github anyways. No need to worry about him. Fixed forms following your suggestion

taxi/forms.py
NEkropo1: Sometimes it's even better not to specify, so malevolent person won't break in.
Consider how easy would be to understand each particular reason, why fake license isn't going inside. And how hard it gonna be to get right point, if you don't have that license and can't simply recheck what you typed.

taxi/forms.py
```diff
@@ -0,0 +1,52 @@
...
+from django.forms import ModelForm
+
+from taxi.models import Car, Driver
+
+
+class DriverLicenseUpdateForm(ModelForm):
+    LICENSE_REQUIRED_LENGTH = 8
+
+    class Meta:
+        model = Driver
```
NEkropo1: checklist

taxi/forms.py
```diff
@@ -0,0 +1,52 @@
...
+        fields = ("license_number",)
+
+    def clean_license_number(self):
+        license_number = self.cleaned_data["license_number"]
+        pattern = r"^[A-Z]{3}[0-9]{5}"
```
NEkropo1: small hint - never use re, if you can, since it's slowing down your python code.

taxi/views.py
```diff
@@ -1,14 +1,16 @@
...
+from .forms import CarForm, DriverLicenseUpdateForm, DriverCreationForm
 from .models import Driver, Car, Manufacturer
 
 
 @login_required
-def index(request):
+def index(request: HttpRequest) -> HttpResponse:
     """View function for the home page of the site."""
 
     num_drivers = Driver.objects.count()
```
NEkropo1: checklist.

taxi/views.py
```diff
@@ -28,6 +30,22 @@ def index(request):
...
 
 
+@login_required()
+def car_driver_assignment(
+        request: HttpRequest,
+        pk: int,
+        driver_id: int
+) -> HttpResponse:
+    car = get_object_or_404(Car, pk=pk)
+    driver = get_object_or_404(Driver, pk=driver_id)
```
NEkropo1: you sure we need to check if we can get object with login_required decorator?)
also checklist.

taxi/views.py
```diff
@@ -28,6 +30,22 @@ def index(request):
...
+
+    if driver in car.drivers.all()[:]:
+        car.drivers.remove(driver_id)
+    else:
+        car.drivers.add(driver_id)
+    return HttpResponseRedirect(reverse_lazy("taxi:car-detail", args=[pk]))
```
NEkropo1: you need that copy here?

taxi/views.py
```diff
@@ -87,3 +105,19 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+
+class DriverUpdateView(generic.UpdateView):
+    model = Driver
+    form_class = DriverLicenseUpdateForm
+
+
+class DriverDeleteView(generic.DeleteView):
+    model = Driver
+    success_url = reverse_lazy("taxi:driver-list")
+    template_name = "taxi/driver_confirmation_delete.html"
```
NEkropo1: Checklist.

NEkropo1 (changes requested):

taxi/forms.py
LiudmylaKulinchenko: Could u fix that, please?

taxi/views.py
```diff
@@ -80,10 +99,27 @@ class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
...
-    model = Driver
+    model = get_user_model()
     paginate_by = 5
 
 
 class DriverDetailView(LoginRequiredMixin, generic.DetailView):
-    model = Driver
-    queryset = Driver.objects.all().prefetch_related("cars__manufacturer")
+    model = get_user_model()
+    queryset = (get_user_model().objects.all()
```
LiudmylaKulinchenko: No need to use `all()` here

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Great job ❄

PR #411, opened 2023-11-20
py-taxi-service-user-form-and-validation Solution

RVChornyy:
Screenshot_1
Screenshot_2
Screenshot_3
Screenshot_4
Screenshot_5
Screenshot_6
Screenshot_7


Sel-Fisher (approved):
very cool

taxi/forms.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    class Meta:
+        model = Driver
+        fields = ("license_number",)
+
+    def clean_license_number(self):
+        license_number = self.cleaned_data["license_number"]
+        return license_number_validator(license_number)
```
wQuelS: Consider implementing something like mixin and add this logic to it, so needed forms can inherit it and receive this functionality without code duplication

taxi/forms.py
```diff
@@ -0,0 +1,41 @@
...
+from django.contrib.auth import get_user_model
+from django.contrib.auth.forms import UserCreationForm
+
+from taxi.models import Driver, Car
+from taxi.validators import license_number_validator
+
+
+class DriverCreateForm(UserCreationForm):
+    class Meta(UserCreationForm.Meta):
+        model = Driver
```
wQuelS: it is better to use `get_user_model()` instead of hardcoding it since your `Driver` is actually a `User`.

P.S. please, always read checklist which may contain some hints, like this one

taxi/forms.py
```diff
@@ -0,0 +1,41 @@
...
+            "license_number",)
+
+    def clean_license_number(self):
+        license_number = self.cleaned_data["license_number"]
+        return license_number_validator(license_number)
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    class Meta:
+        model = Driver
```
wQuelS: same

taxi/validators.py
```diff
@@ -0,0 +1,16 @@
...
+    if not license_number[:3].isupper() or not license_number[:3].isalpha():
+        raise ValidationError(
+            "First 3 characters of your driver license"
+            " must be uppercase letters"
+        )
+    if not license_number[-5:].isdigit():
+        raise ValidationError(
+            "Last 5 characters of your driver license must be digits"
+        )
+    return license_number
```
wQuelS: Using "magic numbers" is not a good practice. Consider adding some constants that will hold them, so it will be also clear for those who may read your code

taxi/views.py
```diff
@@ -87,3 +89,31 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+class DriverLicenseUpdateView(LoginRequiredMixin, generic.UpdateView):
+    model = Driver
+    form_class = DriverLicenseUpdateForm
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+@login_required()
+def car_drivers_setup(request: HttpRequest, pk: int) -> HttpResponse:
+    driver = request.user
+    car_drivers = Car.objects.get(pk=pk).drivers.all()
```
wQuelS: I guess instead of `driver` it would be better to call this variable as a `user` or smth like that, because now it is a bit confusing to have driver and car_drivers:)

taxi/views.py
```diff
@@ -87,3 +89,31 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    if driver in car_drivers:
+        Car.objects.get(pk=pk).drivers.remove(driver)
```
wQuelS: You've already fetched all drivers and stored 'em in a variable. Why didn't you use it here?
It would not only reduce the amount of code but also the number of db hits.

taxi/views.py
```diff
@@ -87,3 +89,31 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    else:
+        Car.objects.get(pk=pk).drivers.add(driver)
+    return redirect(f"/cars/{pk}")
```
wQuelS, line 119: What was the idea behind the decision to use a hardcoded path instead of dynamic routing?

templates/taxi/car_detail.html
```diff
@@ -12,7 +12,13 @@ <h1>
     </a>
   </h1>
   <p>Manufacturer: ({{ car.manufacturer.name }}, {{ car.manufacturer.country }})</p>
-  <h1>Drivers</h1>
+  <h1>Drivers
+  {% if user in car.drivers.all %}
```
wQuelS: here you are triggering a db call for all drivers. Nothing wrong yet, but it can be optimized a bit. 
In most cases, the quantity of drivers ( users ) for the specific car is much bigger than the number of cars that the average user has. 
So, instead of retrieving all drivers for the car, you may check if this vehicle is listed among user's cars.

templates/taxi/driver_confirm_delete.html
```diff
@@ -0,0 +1,11 @@
+{% extends "base.html" %}
+
+{% block content %}
+  <h1>Delete Driver?</h1>
```
wQuelS: Consider adding info what driver is gonna be deleted to make things more clear for the user

wQuelS (changes requested):

taxi/forms.py
```diff
@@ -0,0 +1,55 @@
...
+        return license_number
+
+
+class DriverCreateForm(DriverLicenseValidationMixin, UserCreationForm):
+    class Meta(UserCreationForm.Meta):
+        model = get_user_model()
+        fields = UserCreationForm.Meta.fields + (
+            "first_name",
+            "last_name",
+            "license_number",)
```
LiudmylaKulinchenko: bad formatting

taxi/views.py
LiudmylaKulinchenko: fix that

LiudmylaKulinchenko (changes requested):

taxi/views.py
RVChornyy: i had allready fixed. I changed the if statement: depending on if the exact car is in the list of user's cars, add or remove user from car many-to-many field. Seems like i do not understand how to refactor this function in another way.

wQuelS (approved):
GJ

PR #416, opened 2023-11-20
Solution

D1st3f:


MiskoRuslan (approved):
well done, but in checklist said, that we should use get_user_model() instead Driver, but i think that`s okay

taxi/forms.py
```diff
@@ -0,0 +1,31 @@
+from django.contrib.auth.forms import UserCreationForm
+from django.forms import ModelForm
+
+from taxi.models import Driver
+from taxi.validators import license_number_validator
+
+
+class DriverLicenseUpdateForm(ModelForm):
+    class Meta:
+        model = Driver
```
NEkropo1, line 10: always check if your code is accordingly to checklist.

taxi/forms.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def clean_license_number(self):
+        license_number = self.cleaned_data["license_number"]
+        license_number_validator(license_number)
+        return license_number
+
+
+class DriverCreateForm(UserCreationForm):
+    class Meta:
+        model = Driver
```
NEkropo1, line 21: checklist

taxi/urls.py
```diff
@@ -9,10 +9,14 @@
...
     DriverDetailView,
+    DriverCreateView,
+    DriverUpdateForm,
+    DriverDeleteForm,
+    DriverLicenseUpdateView,
     ManufacturerListView,
     ManufacturerCreateView,
     ManufacturerUpdateView,
-    ManufacturerDeleteView,
+    ManufacturerDeleteView, car_driver_assignment,
```
NEkropo1: fix imports.

taxi/validators.py
```diff
@@ -0,0 +1,18 @@
...
+
+    if not (license_number[:3].isupper() and license_number[:3].isalpha()):
+        raise ValidationError(
+            "License number must start with 3 uppercase letters"
+        )
+
+    if not license_number[3:].isdigit():
+        raise ValidationError(
+            "License number must end with 5 digits"
+        )
```
NEkropo1: Great idea! Consider name it form_validators, since it's only used with forms, not with whole django thing, like i'ts made if we use built-in validators.

taxi/views.py
```diff
@@ -28,6 +30,22 @@ def index(request):
...
 
 
+@login_required()
+def car_driver_assignment(
+        request: HttpRequest,
+        pk: int,
+        driver_id: int
+) -> HttpResponse:
+    car = get_object_or_404(Car, pk=pk)
+    driver = get_object_or_404(Driver, pk=driver_id)
```
NEkropo1: since we got login_required decorator, do we need get_object_or_404?, also checklist )

taxi/views.py
```diff
@@ -28,6 +30,22 @@ def index(request):
...
+
+    if driver in car.drivers.all()[:]:
```
NEkropo1: Why do you need to make a copy of drivers?

taxi/views.py
```diff
@@ -87,3 +105,35 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+
+
+class DriverDeleteForm(LoginRequiredMixin, generic.DeleteView):
+    model = Driver
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+class DriverLicenseUpdateView(LoginRequiredMixin, generic.UpdateView):
+    model = Driver
+    form_class = DriverLicenseUpdateForm
```
NEkropo1: checklist

NEkropo1 (changes requested):
It's obligatory to make code accordingly to checklist. If you see something not as mentioned in that file, you should always request changes)

LiudmylaKulinchenko (approved):
Good job ☀

PR #417, opened 2023-11-21
Solution

sakhaline:
Screenshot from 2023-11-29 04-55-34
Screenshot from 2023-11-29 04-56-39
Screenshot from 2023-11-29 04-56-55
Screenshot from 2023-11-29 04-57-08
Screenshot from 2023-11-29 04-57-17
Screenshot from 2023-11-29 04-57-31
Screenshot from 2023-11-29 04-57-45
Screenshot from 2023-11-29 04-57-54
Screenshot from 2023-11-29 04-57-58
Screenshot from 2023-11-29 04-58-25
Screenshot from 2023-11-29 04-58-30
Screenshot from 2023-11-29 04-58-37
Screenshot from 2023-11-29 04-58-48
Screenshot from 2023-11-29 04-58-56
Screenshot from 2023-11-29 04-59-04
Screenshot from 2023-11-29 04-59-11
Screenshot from 2023-11-29 04-59-19
Screenshot from 2023-11-29 05-07-53


DenPrislipskyi (approved):

taxi/forms.py
```diff
@@ -0,0 +1,37 @@
+from django import forms
+from django.contrib.auth import get_user_model
+from django.contrib.auth.forms import UserCreationForm
+
+from taxi.models import Driver, Car
+from validations.license_validation import validate_license_number
+
+
+class DriverForm(UserCreationForm):
```
wQuelS: I guess it would be better to name it `DriverCreationForm` here

taxi/forms.py
```diff
@@ -0,0 +1,37 @@
...
+    class Meta:
+        model = get_user_model()
+        fields = UserCreationForm.Meta.fields + ("license_number",)
```
wQuelS: why have you decided not to add `first_name`, and `second_name` to the creation form?

taxi/forms.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def clean_license_number(self):
+        license_number = self.cleaned_data["license_number"]
+        return validate_license_number(license_number)
+
```
wQuelS: this method is duplicated. Consider implementing something like mixin, so both forms will just inherit it and get needed functionality without code duplication

taxi/forms.py
```diff
@@ -0,0 +1,37 @@
...
+
+class CarForm(forms.ModelForm):
+    drivers = forms.ModelMultipleChoiceField(
+        queryset=get_user_model().objects.all(),
+        widget=forms.CheckboxSelectMultiple,
+    )
```
wQuelS: I bet it should be possible to create a car without specifying a driver but adding him later, when needed. Is it possible with your approach?

taxi/urls.py
```diff
@@ -46,6 +51,22 @@
...
+    path("drivers/create/", DriverCreateView.as_view(), name="driver-create"),
+    path(
+        "drivers/<int:pk>/license-update/",
+        DriverLicenseUpdateView.as_view(),
+        name="driver-update"
+    ),
+    path(
+        "drivers/<int:pk>/update",
+        DriverUpdateView.as_view(),
+        name="driver-full-update"
```
wQuelS: why have you decided to have separate endpoints for license updates and general updates? Do they somehow differ in your implementation?

taxi/urls.py
wQuelS, line 1: am I blind or you have forgotten to add endpoints for assigning/removing drivers to/from cars?

taxi/urls.py
```diff
@@ -46,6 +51,22 @@
...
         "drivers/<int:pk>/", DriverDetailView.as_view(), name="driver-detail"
     ),
+    path("drivers/create/", DriverCreateView.as_view(), name="driver-create"),
+    path(
+        "drivers/<int:pk>/license-update/",
+        DriverLicenseUpdateView.as_view(),
+        name="driver-update"
+    ),
+    path(
+        "drivers/<int:pk>/update",
```
wQuelS: seems like you're missing trailing `/` here!?:)

taxi/views.py
```diff
@@ -63,8 +66,8 @@ class CarDetailView(LoginRequiredMixin, generic.DetailView):
 
 
 class CarCreateView(LoginRequiredMixin, generic.CreateView):
+    form_class = CarForm
```
wQuelS: what about same form but for `CarUpdateView`?

taxi/views.py
```diff
@@ -87,3 +90,41 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+
+class DriverCreateView(LoginRequiredMixin, generic.CreateView):
+    model = Driver
+    form_class = DriverForm
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+class DriverDeleteView(LoginRequiredMixin, generic.DeleteView):
+    model = Driver
+    template_name = "taxi/driver_confirm_delete.html"
```
wQuelS: your template naming is similar to default, so Django will automatically find needed template and this attr is redundant, I guess

taxi/views.py
```diff
@@ -87,3 +90,41 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+class DriverUpdateView(generic.UpdateView):
+    model = get_user_model()
```
wQuelS, line 129: here u are using `get_user_model()`. Why not use it in previous views? Fix somewhere to make your code more consistent.

Also, why is it possible to update the driver's data without login?

taxi/views.py
```diff
@@ -87,3 +90,41 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    form_class = DriverLicenseUpdateForm
+    template_name = "taxi/license_form.html"
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+@login_required()
+def car_drivers_view(request: HttpRequest, pk: int) -> HttpResponse:
+    drivers = Car.objects.get(pk=pk).drivers
+
+    if request.user in drivers.all():
```
wQuelS: this might be very inefficient if you have a lot of drivers. Just to assign/remove a driver to/from a car, you'd make this db hit each time.

templates/taxi/car_confirm_delete.html
wQuelS, line 1: It should be possible to return to the previous page if I, for example, decided not to delete a car. Add something like "cancel" button for each delete_confirmation page

templates/taxi/car_detail.html
```diff
@@ -13,10 +13,16 @@ <h1>
   </h1>
   <p>Manufacturer: ({{ car.manufacturer.name }}, {{ car.manufacturer.country }})</p>
   <h1>Drivers</h1>
+  {% if user in car.drivers.all %}
+    <a class="btn btn-primary" href="{% url 'taxi:car-driver-update' pk=car.id %}">Assign me to this car</a>
+  {% else %}
+    <a class="btn btn-primary" href="{% url 'taxi:car-driver-delete' pk=car.id %}">Delete me from this car"></a>
+  {% endif %}
```
wQuelS: so, if a user is already listed as a driver for some car, he will see `Assign me to this car`? Feels weird

templates/taxi/driver_confirm_delete.html
```diff
@@ -0,0 +1,10 @@
+{% extends "base.html" %}
+{% load crispy_forms_filters %}
+
+{% block content %}
+  <h1>Are you sure you want to delete driver?</h1>
+  <form action="" method="post">
+    {% csrf_token %}
+    <input class="btn btn-danger" type="submit" value="Yes">
```
wQuelS: As I mentioned earlier, add "cancel" button which will lead to previous page

templates/taxi/driver_form.html
wQuelS, line 1: You can use this template with `object|yesno:"Update,Create"`

validations/license_validation.py
```diff
@@ -0,0 +1,18 @@
+from django.core.exceptions import ValidationError
+
+
+def validate_license_number(license_number):
+    first_three_characters = license_number[:3]
+    last_five_characters = license_number[-5:]
+
+    if len(license_number) != 8:
```
wQuelS: some of those "magic numbers" are better to define as constants, so it will be clear for everyone what do they mean:)

wQuelS (changes requested):

wQuelS (commented):
Also, add screenshots to all pages, including delete confirmation and so on

taxi/urls.py
sakhaline: yes, they differ, since the URLs lead to different views, which in turn use different forms. The license update page contains only the license number field, and the driver update page contains all the fields that are associated with it

sakhaline: Agree, this is redundant

taxi/urls.py
sakhaline: I implemented this feature using a class just to practice a different approach.
I'm just getting ahead of the question, "why did you decide to do this?" ;)

wQuelS (approved):
GJ

PR #418, opened 2023-11-21
Validators and updates

Sel-Fisher:


Sel-Fisher:
image
image
image
image
image
image

Wonsky1 (approved):

taxi/forms.py
```diff
@@ -0,0 +1,57 @@
...
+
+class DriverLicenseUpdateForm (forms.ModelForm):
+    class Meta:
+        model = Driver
+        fields = ("license_number",)
+
+    def clean_license_number(self):
+        license_number = self.cleaned_data["license_number"]
+        validator_license_number(license_number)
+        return license_number
```
wQuelS: this method is just a duplicate of the `DriverCreationForm` method. Maybe implement some mixin, so both classes can inherit it and use this logic without copy paste?

taxi/urls.py
```diff
@@ -42,10 +46,24 @@
     path("cars/create/", CarCreateView.as_view(), name="car-create"),
     path("cars/<int:pk>/update/", CarUpdateView.as_view(), name="car-update"),
     path("cars/<int:pk>/delete/", CarDeleteView.as_view(), name="car-delete"),
+    path("cars/<int:pk>/assign-me/", assing_me, name="car-assign"),
```
wQuelS: Since this endpoint works to assign and remove users from cars, the name should be more clear. Because now it is possible to think that it is possible to assign someone to a car, but not remove

taxi/views.py
```diff
@@ -64,7 +77,7 @@ class CarDetailView(LoginRequiredMixin, generic.DetailView):
 
 class CarCreateView(LoginRequiredMixin, generic.CreateView):
     model = Car
-    fields = "__all__"
+    form_class = CarForm
```
wQuelS, line 80: what about `form_class` for `CarUpdateView`?

taxi/views.py
```diff
@@ -87,3 +100,20 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+
+
+class DriverCreateView(LoginRequiredMixin, generic.CreateView):
+    model = Driver
+    form_class = DriverCreationForm
+
+
+class DriverDeleteView(LoginRequiredMixin, generic.DeleteView):
+    model = Driver
+    template_name = "taxi/driver_confirm_delete.html"
```
wQuelS: I guess it will automatically look for this template and there is no need to explicitly set it here

taxi/views.py
```diff
@@ -28,6 +30,17 @@ def index(request):
     return render(request, "taxi/index.html", context=context)
 
 
+@login_required
+def assing_me(request: HttpRequest, pk: int) -> HttpResponse:
+    drivers = Car.objects.get(pk=pk).drivers
+    user = request.user
+    if request.user in drivers.all():
```
wQuelS, line 37: this might be very inefficient if there are a lot of drivers in your system

templates/taxi/driver_confirm_delete.html
```diff
@@ -0,0 +1,11 @@
...
+
+{% block content %}
+  <h1>Delete Driver {{ driver.username }}?</h1>
+  <p>All info about this user will be deleted!</p>
+  <form action="" method="post">
+    {% csrf_token %}
+
+    <input type="submit" value="Yes" class="btn btn-danger">
+  </form>
+{% endblock %}
```
wQuelS: How to move back if you change your mind to delete a driver?

templates/taxi/driver_detail.html
```diff
@@ -3,6 +3,9 @@
 {% block content %}
   <h1>
     Username: {{ driver.username }}
+    <a href="{% url 'taxi:driver-update' pk=driver.id %}" class="btn btn-primary link-to-page">
+      Update license
+    </a>
```
wQuelS, line 11: what about `delete`? Isn't it logically correct to place such a button on a driver's detail page?

templates/taxi/driver_form.html
```diff
@@ -0,0 +1,12 @@
...
+
+{% block content %}
+  <h1>Create Driver</h1>
+  <form action="" method="post" novalidate>
+    {% csrf_token %}
+    {{ form|crispy }}
+
+    <input type="submit" value="Submit" class="btn btn-primary">
+  </form>
+{% endblock %}
```
wQuelS: why not `object|yesno:"Update,Create"`?

wQuelS (changes requested):

wQuelS (changes requested):
Left some comments above

taxi/views.py
Sel-Fisher: missed up

taxi/urls.py
Sel-Fisher: will overthink it

taxi/forms.py
Sel-Fisher: good notice

taxi/views.py
Sel-Fisher: missed up

templates/taxi/driver_confirm_delete.html
Sel-Fisher: missed up

templates/taxi/driver_detail.html
Sel-Fisher: missed up

templates/taxi/driver_form.html
Sel-Fisher: forgot

LiudmylaKulinchenko (approved):
Good job 🦋

PR #421, opened 2023-11-21
Taxi service user form and validation solution

Lyutillis:


Lyutillis:
Create Driver:
image
Update Driver License:
image
Delete Driver:
image
Car Create:
image
Car Detail Page:
image
image

sakhaline (approved):
GJ

taxi/forms.py
```diff
@@ -0,0 +1,37 @@
...
+        validators=[
+            RegexValidator(regex="^[A-Z]{3}\\d{5}$",
+                           message=("License number must be of length 8,"
+                                    " start with 3 uppercase letters"
+                                    " and end with 5 digits"))
+        ]
+    )
+
+    class Meta(UserCreationForm.Meta):
+        model = Driver
```
NEkropo1, line 23: Always check if your code is according to checklist.

taxi/forms.py
```diff
@@ -0,0 +1,37 @@
...
+        validators=[
+            RegexValidator(regex="^[A-Z]{3}\\d{5}$",
+                           message=("License number must be of length 8,"
+                                    " start with 3 uppercase letters"
+                                    " and end with 5 digits"))
+        ]
+    )
+
+    class Meta:
+        model = Driver
```
NEkropo1, line 32: checklist

taxi/forms.py
```diff
@@ -0,0 +1,37 @@
...
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    license_number = forms.CharField(
+        required=True,
+        validators=[
+            RegexValidator(regex="^[A-Z]{3}\\d{5}$",
+                           message=("License number must be of length 8,"
+                                    " start with 3 uppercase letters"
+                                    " and end with 5 digits"))
+        ]
```
NEkropo1: That's good, but a small hint - always try not to use regex where you can not use it since it's much slower and not that efficient and readable as native python code.

taxi/views.py
```diff
@@ -28,6 +31,22 @@ def index(request):
...
+    car.drivers.add(user)
+    return redirect("taxi:car-detail", pk=pk)
+
+
+@login_required
+def car_unassign(request, pk: int):
+    user = request.user
+    car = Car.objects.get(id=pk)
+    car.drivers.remove(user)
+    return redirect("taxi:car-detail", pk=pk)
```
NEkropo1: You can make that as one function and change according template as it was in video. Not obligatory, good job, but think about it.

taxi/views.py
```diff
@@ -28,6 +31,22 @@ def index(request):
...
+
+
 class ManufacturerListView(LoginRequiredMixin, generic.ListView):
     model = Manufacturer
     context_object_name = "manufacturer_list"
```
NEkropo1: you don't need to add context_object_name if that's default value. You can read more here: [django class based views](https://docs.djangoproject.com/en/4.2/topics/class-based-views/generic-display/)

taxi/views.py
```diff
@@ -67,6 +86,17 @@ class CarCreateView(LoginRequiredMixin, generic.CreateView):
...
+        if not form_class:
+            form_class = self.get_form_class()
+        form = super().get_form(form_class)
+        form.fields["drivers"] = forms.ModelMultipleChoiceField(
+            queryset=Driver.objects.all(),
+            widget=forms.CheckboxSelectMultiple,
+            required=True,
+        )
+        return form
+
```
NEkropo1: For better separation of concerns and adhering to the principles of modularity and reusability, modifying the form directly rather than within the view is a preferable approach.
I meant simply add this field inside form, not inside view, that seems more logical

taxi/views.py
```diff
@@ -80,10 +110,27 @@ class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
...
-    model = Driver
+    model = get_user_model()
     paginate_by = 5
 
 
 class DriverDetailView(LoginRequiredMixin, generic.DetailView):
-    model = Driver
+    model = get_user_model()
     queryset = Driver.objects.all().prefetch_related("cars__manufacturer")
+
```
NEkropo1: While the 'model' attribute is set to 'get_user_model()', indicating a change in the expected model for the view, it's crucial to note that the 'queryset' attribute directly references 'Driver.objects.all()' without incorporating the model derived from 'get_user_model()'. As a result, the 'queryset' doesn't align with the updated 'model'.

templates/taxi/car_detail.html
```diff
@@ -11,6 +11,15 @@ <h1>
...
   </h1>
+  {% if user not in car.drivers.all %}
+  <a href="{% url 'taxi:car-assign' pk=car.id %}" class="btn btn-success link-to-page">
+    Assign me to this car
+  </a>
+  {% else %}
+  <a href="{% url 'taxi:car-unassign' pk=car.id %}" class="btn btn-danger link-to-page">
+    Delete me from this car
+  </a>
+  {% endif %}
```
NEkropo1: remember what I said about this function?
mentioned for ease.

NEkropo1 (changes requested):

LiudmylaKulinchenko (approved):
Nice 🔥

PR #472, opened 2024-01-27
Solution

Lebwa1337:
image
image
image
image
image
image
image
image
image
image


Lebwa1337:
Sorry for all that commits,  I've been having authorization issues on git hub for the last 2 days, so these commits are me testing some stuff

taxi/migrations/0001_initial.py
```diff
@@ -1,66 +1,181 @@
...
+                (
+                    "manufacturer",
+                    models.ForeignKey(
+                        on_delete=django.db.models.deletion.CASCADE,
+                        to="taxi.manufacturer",
+                    ),
+                ),
             ],
         ),
     ]
```
Dimosphen1: Do not overwrite the initial migration, create a new migration apart from the existing one.

requirements.txt
Dimosphen1, line 1: Do not change requirements other than adding `crispy-bootstrap4`, there is no need to change the present library versions.

Dimosphen1 (changes requested):
Several changes were requested.

taxi/migrations/0001_initial.py
Lebwa1337: But initial migration doesn`t working all the time.When i`m trying to run another migration nothing happens.So,i need to delete this migration and run new one, then all works

requirements.txt
Lebwa1337: i used pip freeze, so that`s why it add all these to requirements

taxi/migrations/0001_initial.py
viktoria-rybenchuk: Which command do you use to run initial migration? You should just use `python manage.py migrate`

viktoria-rybenchuk (changes requested):
not fixed

viktoria-rybenchuk (approved):
Good job!

PR #492, opened 2024-02-23
Solution

ArturPoltser:
image
image
image
image
image
image

### DIFF
--- a/taxi/models.py
+++ b/taxi/models.py
@@ -1,3 +1,4 @@
+from django.core.validators import RegexValidator
 from django.db import models
 from django.contrib.auth.models import AbstractUser
 from django.urls import reverse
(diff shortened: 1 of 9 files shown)


kabust (approved):

taxi/models.py
```diff
@@ -15,7 +16,18 @@ def __str__(self):
 
 
 class Driver(AbstractUser):
-    license_number = models.CharField(max_length=255, unique=True)
+    license_number = models.CharField(
+        max_length=255,
+        unique=True,
+        validators=[
+            RegexValidator(
+                regex=r"^[A-Z]{3}\d{5}$",
```
NEkropo1: ok, but try not to use regex, since it's too complex for anyone who doesn't know exactly how it works to check and change and it's not explicit, also, sometimes, too slow.
Also, better not to add such functionality on models, since you won't be able to make direct changes to db via admin panel, etc, better use form and form_validation.

NEkropo1 (changes requested):

NEkropo1 (approved):
Great job!

PR #494, opened 2024-02-23
Solution

kabust:
Driver detailed

image

Driver update:

image

Driver delete:

image

Car detailed:

image

Driver create:

image


taxi/forms.py
```diff
@@ -0,0 +1,59 @@
...
+        model = Driver
+        fields = (
+            "username",
+            "password1",
+            "password2",
+            "license_number",
+            "email",
+            "first_name",
+            "last_name"
+        )
```
aLEKS-e3: why don't u select fields like this? `fields = UserCreationForm.Meta.fields + ("custom_field",)`

taxi/views.py
```diff
@@ -61,10 +64,13 @@ class CarListView(LoginRequiredMixin, generic.ListView):
 class CarDetailView(LoginRequiredMixin, generic.DetailView):
     model = Car
 
+    def post(self, pk: int):
+        print("post sent")
+
```
aLEKS-e3: why do you need this?

templates/taxi/driver_confirm_delete.html
```diff
@@ -0,0 +1,9 @@
+{% extends "base.html" %}
+
+{% block content %}
+  <h1>Delete driver {{ driver.username }}?</h1>
+  <form action="" method="post">
+    {% csrf_token %}
+    <input type="submit" value="Yes" class="btn btn-danger">
+  </form>
+{% endblock %}
```
aLEKS-e3: add a blank line in the end)

aLEKS-e3 (changes requested):

taxi/forms.py
kabust: do we need some of the default fields? i wanted to get rid of some of the unnecessary ones, such as superuser status, permissions, etc

taxi/views.py
kabust: forgot to remove, thx

aLEKS-e3 (approved):
nice!

taxi/forms.py
NEkropo1: Also you can provide fields in wanted ordere, while using meta they always will be in meta order.

NEkropo1 (approved):
Good job!

PR #498, opened 2024-02-27
Solution

MaxymChyncha:
Driver List Page
image
Driver Create Page
image
Driver Detail Page
image
Driver Update License Page
image
Driver Delete Page
image
Car Create Page
image
Car Detail Page before Assignment
image
Car Detail Page after Assingment
image


barabarinov (approved):
LGTM!

taxi/models.py
```diff
@@ -15,7 +16,21 @@ def __str__(self):
...
+        unique=True,
+        validators=[
+            validators.RegexValidator(
+                regex="^[A-Z]{3}\\d{5}$",
+                message=(
+                    "Your license number should "
+                    "start with 3 uppercase letters "
+                    "and end with 5 digits."
+                ),
+                code="wrong license number"
```
NEkropo1: Move this check to view validation via form, since you won't be able to create custom user license via admin panel. Also better not to use regex since it's not intuitive and explicit also, sometimes, pretty slow.

NEkropo1 (changes requested):

NEkropo1 (approved):
Good job!

PR #534, opened 2024-03-18
solved task

Molchanov-Ihor:


taxi/forms.py
```diff
@@ -0,0 +1,59 @@
...
+
+def validate_license_number(license_number):
+    license_number_size = 8
+    seria_size = 3
+    if len(license_number) != license_number_size:
+        raise ValidationError(
+            f"License number must be at {license_number_size} chars"
+        )
+    seria = license_number[:seria_size]
+    number = license_number[seria_size:]
```
viktoria-rybenchuk: no need to define variables, you can directly make checking

taxi/urls.py
```diff
@@ -13,6 +12,9 @@
     ManufacturerCreateView,
     ManufacturerUpdateView,
     ManufacturerDeleteView,
+    DriverUpdateLicenseNumberView,
+    DriverCreateView,
+    DriverDeleteView, car_detail,
```
viktoria-rybenchuk: fix imports

taxi/views.py
```diff
@@ -58,13 +66,28 @@ class CarListView(LoginRequiredMixin, generic.ListView):
...
+        "car": car
+    }
+    if request.method == "GET":
+        return render(request, "taxi/car_detail.html", context=context)
+
+    if request.method == "POST":
+        driver = get_user_model().objects.get(pk=request.user.pk)
+        if "assign" in request.POST:
+            car.drivers.add(driver)
+        if "remove" in request.POST:
```
viktoria-rybenchuk: you can check if request user already in car.drivers and then remove it, otherwise add

taxi/views.py
```diff
@@ -84,6 +107,37 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    def form_valid(self, form):
+        car_id = self.kwargs["pk"]
+        car = Car.objects.get(pk=car_id)
+        driver = Driver.objects.get(pk=self.request.user.pk)
+
+        if "assign" in self.request.POST:
+            car.drivers.add(driver)
+        if "remove" in self.request.POST:
+            car.drivers.remove(driver)
+        return super().form_valid(form)
```
viktoria-rybenchuk: why do you need the same logic as you have defined above?

viktoria-rybenchuk (changes requested):

taxi/views.py
Molchanov-Ihor: forgot to delete the old version, sorry

viktoria-rybenchuk (approved):
Well done!

PR #545, opened 2024-03-26
solution

VolodymyrSeniv:
image
image
image
image
image
image
image


taxi/views.py
```diff
@@ -79,6 +83,28 @@ class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
     success_url = reverse_lazy("taxi:car-list")
 
 
+@login_required
+def user_add_to_car(request: HttpRequest, pk) -> HttpResponse:
```
vsmutok: use class with post method instead of function view

taxi/views.py
```diff
@@ -79,6 +83,28 @@ class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
...
+    car.drivers.add(request.user)
+    context = {
+        "car": car,
+        "pk": pk
+    }
+    return render(request, "taxi/car_detail.html", context=context)
+
+
+@login_required
+def user_delete_from_car(request: HttpRequest, pk) -> HttpResponse:
```
vsmutok: use class with post method instead of function view

staticfiles/admin/css/autocomplete.css
vsmutok, line 1: why you need all this staticfiles?
plese remove them from PR

vsmutok (changes requested):

EdAlekseiev (approved):
Good job

PR #582, opened 2024-05-12
solution

panicua:
Checkboxes:
chrome_EYsr7Vnbwr

Before pressing "Delete me from this car":
chrome_ThC2jzzrwr

After pressing "Delete me from this car":
chrome_GQjv9z7U97

Page where you can Delete driver and also Cancel detele > redirect to drivers-list:
chrome_3cxPtrkOnl

Driver list page with "Create" button that needed to be created:
chrome_YppT903YHg

Create driver with one of the validators sending message:
chrome_lgkqWZQ1jD

Update/Delete Buttons on driver-detail page:
chrome_N1pVFfPWIE

Update page with other validator sending a message:
chrome_nshvGsb1rI


taxi/views.py
```diff
@@ -87,3 +89,36 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    return redirect("taxi:car-detail", pk=car_id)
+
+
+def remove_driver_from_car(request):
+    car_id = request.POST.get("car_id")
+    driver = request.user
+
+    Car.objects.get(pk=car_id).drivers.remove(driver)
+
+    return redirect("taxi:car-detail", pk=car_id)
```
YuriiKindrat: it looks pretty similar, could u make it with only one view ?

YuriiKindrat (changes requested):

taxi/views.py
panicua: but don't we break the principle of singe responsibility ?

taxi/views.py
```diff
@@ -87,3 +90,33 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
 class DriverDetailView(LoginRequiredMixin, generic.DetailView):
     model = Driver
     queryset = Driver.objects.all().prefetch_related("cars__manufacturer")
```
MNDonut: all is redundant

taxi/views.py
```diff
@@ -87,3 +90,33 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    action = request.POST.get("action")
+
+    if action == "add":
+        Car.objects.get(pk=car_id).drivers.add(driver)
+    elif action == "remove":
+        Car.objects.get(pk=car_id).drivers.remove(driver)
+    else:
+        return HttpResponse("Invalid action", status=400)
+
+    return redirect("taxi:car-detail", pk=car_id)
```
MNDonut: let's make it view as well, not function

MNDonut (changes requested):

taxi/views.py
MNDonut: @panicua nope, we don't because REST says that one url e.g. users/ can support GET(list) and POST methods and users/{id}/ can support get, patch, put delete. In this case we don't break that principle

taxi/views.py
panicua: This code was there before I added my code, I don't think we're supposed to write a code from start, nothing in this task says otherwise.

Perhaps I don't understand what you ment by that comment.

update:
oh.... you mean all(), I thought "all" 3 lines of code are redundant. But still, these code wasn't written by me, it was there before me, need to fix task starting files :)

fsocie7y (approved):
GJ 👍

PR #589, opened 2024-05-13
Solution

Chub-andrew:


taxi/forms.py
```diff
@@ -0,0 +1,55 @@
...
+                "Last five characters should be digits"
+            )
+        return license_number
+
+
+class CarForm(forms.ModelForm):
+    drivers = forms.ModelMultipleChoiceField(
+        queryset=get_user_model().objects.all(),
+        widget=CheckboxSelectMultiple(),
+        required=False)
```
YuriiKindrat: bad formatting here

YuriiKindrat (changes requested):

taxi/urls.py
```diff
@@ -46,6 +50,24 @@
     path(
         "drivers/<int:pk>/", DriverDetailView.as_view(), name="driver-detail"
     ),
+    path(
+        "drivers/create", DriverCreateView.as_view(), name="driver-create"
```
MNDonut: url doesn't end with /

taxi/views.py
```diff
@@ -64,8 +65,9 @@ class CarDetailView(LoginRequiredMixin, generic.DetailView):
 
 class CarCreateView(LoginRequiredMixin, generic.CreateView):
     model = Car
-    fields = "__all__"
+    # fields = "__all__"
```
MNDonut: remove this comment

taxi/views.py
```diff
@@ -87,3 +89,31 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
 class DriverDetailView(LoginRequiredMixin, generic.DetailView):
     model = Driver
     queryset = Driver.objects.all().prefetch_related("cars__manufacturer")
```
MNDonut: all() is redundant, because it calls 2 sql queries: fetch all and then for each record fetch cars__manufacturer. It's better to call prefetch_related directly

taxi/views.py
```diff
@@ -87,3 +89,31 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    model = Driver
+    form_class = DriverLicenseUpdateForm
+    success_url = reverse_lazy("taxi:driver-list")
+    template_name = "taxi/driver_license_update_form.html"
+
+
+def assign_or_delete_car(request, pk):
+    car = get_object_or_404(Car, id=pk)
+    driver = request.user
+    if car in driver.cars.all():
```
MNDonut: it's better to use filter for driver.cars and check if driver owns this car

MNDonut (changes requested):

taxi/views.py
```diff
@@ -112,7 +111,7 @@ class DriverUpdateView(LoginRequiredMixin, generic.UpdateView):
 def assign_or_delete_car(request, pk):
     car = get_object_or_404(Car, id=pk)
     driver = request.user
-    if car in driver.cars.all():
+    if driver.cars.filter(id=pk).exist():
```
MNDonut: queryset doesn't have exist method :)

MNDonut (changes requested):

taxi/views.py
Chub-andrew: I was inattentive, but after checking, everything is working now.

fsocie7y (approved):
Finally, GJ 👍

PR #624, opened 2024-06-22
solution

nrjss67:


taxi/views.py
```diff
@@ -28,6 +31,16 @@ def index(request):
     return render(request, "taxi/index.html", context=context)
 
 
+def car_add_driver(request, pk):
```
vasylhnatiuk, line 34: it is better to use class  view with post method instead of function view

templates/taxi/driver_confirm_delete.html
```diff
@@ -0,0 +1,10 @@
+{% extends "base.html" %}
+
+{% block content %}
+  <h1>Delete driver?</h1>
+  <form action="" method="post">
+    {% csrf_token %}
+
+    <input type="submit" value="Yes" class="btn btn-danger">
+  </form>
+{% endblock %}
```
vasylhnatiuk, line 10: add newline

vasylhnatiuk (changes requested):

PR #640, opened 2024-07-03
Add driver forms and validation

anton-bilous:
image
image
image
image
image
image
image
image
image
image
image
image
image

### DIFF
--- a/taxi/admin.py
+++ b/taxi/admin.py
@@ -1,6 +1,6 @@
 from django.contrib import admin
 from django.contrib.auth.admin import UserAdmin
-from .models import Driver, Car, Manufacturer
+from taxi.models import Driver, Car, Manufacturer
 
 
 @admin.register(Driver)
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -2,7 +2,7 @@ from django import forms
 from django.contrib.auth import get_user_model
 from django.contrib.auth.forms import UserCreationForm
 
-from .models import Car
+from taxi.models import Car
 
 
 class DriverCreationForm(UserCreationForm):
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -1,0 +1,33 @@
+from django import forms
+from django.contrib.auth import get_user_model
+from django.contrib.auth.forms import UserCreationForm
+
+from .models import Car
+
+
+class DriverCreationForm(UserCreationForm):
+    class Meta(UserCreationForm.Meta):
+        model = get_user_model()
+        fields = UserCreationForm.Meta.fields + (
+            "first_name",
+            "last_name",
+            "license_number",
+        )
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    class Meta:
+        model = get_user_model()
+        fields = ("license_number",)
+
+
+class CarForm(forms.ModelForm):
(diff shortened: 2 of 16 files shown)


taxi/forms.py
```diff
@@ -0,0 +1,33 @@
+from django import forms
+from django.contrib.auth import get_user_model
+from django.contrib.auth.forms import UserCreationForm
+
+from .models import Car
```
Oleksl888: avoid using relative imports

taxi/models.py
```diff
@@ -15,7 +16,25 @@ def __str__(self):
 
 
 class Driver(AbstractUser):
-    license_number = models.CharField(max_length=255, unique=True)
+    @staticmethod
+    def validate_license_number(value: str) -> None:
```
Oleksl888: validation needs to be added on forms, try using built-in Validators there to simplify code

taxi/views.py
```diff
@@ -64,13 +66,13 @@ class CarDetailView(LoginRequiredMixin, generic.DetailView):
...
     model = Car
-    fields = "__all__"
+    form_class = CarForm
     success_url = reverse_lazy("taxi:car-list")
 
 
 class CarUpdateView(LoginRequiredMixin, generic.UpdateView):
     model = Car
-    fields = "__all__"
+    form_class = CarForm
```
Oleksl888: there is no requirement to change update form

taxi/views.py
```diff
@@ -87,3 +89,42 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+class DriverLicenseUpdateView(LoginRequiredMixin, generic.UpdateView):
+    model = Driver
+    form_class = DriverLicenseUpdateForm
+
+    def get_success_url(self):
+        return reverse(
+            "taxi:driver-detail", args=(self.get_context_data()["object"].id,)
```
Oleksl888: better use `self.object.id` here

taxi/views.py
```diff
@@ -87,3 +89,42 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+        )
+
+
+@login_required
```
Oleksl888: You can put this logic into `post` method of `CarDetailView`. The reason is that we don't need to create extra views and url endpoints to handle adding/deleting driver.

1. Add form on car detail template instead of href's that will send value 'Add' or 'Delete' to post method to the cardetailview.

templates/taxi/car_detail.html
```diff
@@ -12,7 +12,13 @@ <h1>
     </a>
   </h1>
   <p>Manufacturer: ({{ car.manufacturer.name }}, {{ car.manufacturer.country }})</p>
-  <h1>Drivers</h1>
+  <h1>Drivers
+    {% if user not in car.drivers.all %}
```

Oleksl888 (changes requested):

taxi/admin.py
```diff
@@ -1,6 +1,6 @@
 from django.contrib import admin
 from django.contrib.auth.admin import UserAdmin
-from .models import Driver, Car, Manufacturer
+from taxi.models import Driver, Car, Manufacturer
```
MNDonut, line 3: split different groups of imports

MNDonut (approved):

PR #655, opened 2024-07-15
Solution

SrMusienko:


SrMusienko:
1в
2в
3в
4в
5в
6в
7в

taxi/forms.py
```diff
@@ -0,0 +1,42 @@
+from django import forms
+from django.core.exceptions import ValidationError
+from django.contrib.auth.forms import UserCreationForm
+
+from .models import Driver, Car
```
maxgerman: generally, absolute imports are more reliable and should be preferred. From python docs:
> Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path):

taxi/forms.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def validate_license_number(value):
+    if len(value) != 8:
+        raise ValidationError("License number must have 8 characters!")
+    elif not value[0:3].isalpha():
```
maxgerman: there's no point in using `elif` after `raise`.
Also, slices can be shortened to e.g. just [:3].
These apply to all such cases in this review

taxi/urls.py
```diff
@@ -13,6 +13,11 @@
     ManufacturerCreateView,
     ManufacturerUpdateView,
     ManufacturerDeleteView,
+    DriverCreateView,
+    DriverUpdateView,
+    DriverDeleteView,
+    assign_driver_to_car,
+    delete_driver_from_car,
```
maxgerman: It's not an error but it's a good practice to keep such things with multiple values as imports, maps, etc in alphabetical order, esp in bigger projects. This improves readability and maintenance.
For imports you can use tools like `isort`

taxi/urls.py
```diff
@@ -37,15 +42,64 @@
...
+        DriverUpdateView.as_view(),
+        name="driver-update"
+    ),
+    path(
+        "drivers/<int:pk>/delete/",
+        DriverDeleteView.as_view(),
+        name="driver-delete"
+    ),
+    path(
+        "assign/driver/<int:driver_id>/to/car/<int:car_id>",
```
maxgerman: why do we need so many url parts here? Try to get rid of redundant parts and make these consistent with other endpoints in terms of naming. E.g. first part as `drivers/` seems logical, continue from here

taxi/urls.py
```diff
@@ -37,15 +42,64 @@
...
+        assign_driver_to_car,
+        name="driver-assign-to-car"
+    ),
+    path(
+        "delete/driver/<int:driver_id>/from/car/<int:car_id>/",
```
maxgerman: same as above

taxi/views.py
```diff
@@ -1,9 +1,11 @@
+from django.contrib.auth import get_user_model
 from django.contrib.auth.decorators import login_required
-from django.shortcuts import render
+from django.shortcuts import render, redirect
 from django.urls import reverse_lazy
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
 
+from .forms import DriverCreationForm, DriverLicenseUpdateForm
```
maxgerman: multiple imports - new lines - see the checklist

taxi/views.py
```diff
@@ -28,6 +30,24 @@ def index(request):
     return render(request, "taxi/index.html", context=context)
 
 
+def assign_driver_to_car(request, car_id, driver_id):
+    car = Car.objects.get(id=car_id)
+    driver = Driver.objects.get(id=driver_id)
```
maxgerman: think of what would happen if wrong ids are passed here? What's the more appropriate method to handle this more gracefully?

taxi/views.py
```diff
@@ -28,6 +30,24 @@ def index(request):
...
+
+    car.drivers.add(driver)
+
+    return redirect("taxi:car-detail", pk=car_id)
+
+
+def delete_driver_from_car(request, car_id, driver_id):
+    car = Car.objects.get(id=car_id)
+    driver = Driver.objects.get(id=driver_id)
```
maxgerman: same as above

maxgerman (changes requested):
Good job! Some style considerations need to be adopted, but don't worry, you get used to it :)

maxgerman:
test

templates/taxi/driver_form.html
```diff
@@ -0,0 +1,13 @@
...
+{% block content %}
+  <h1>{{ object|yesno:"Update,Create" }} driver</h1>
+  <form action="" method="post" novalidate>
+    {% csrf_token %}
+    {{ form|crispy }}
+
+    <input type="submit" value="Submit" class="btn btn-primary">
+    <button type="button" class="btn btn-secondary" onclick="window.history.back()">Cancel</button>
+  </form>
+{% endblock %}
```
maxgerman, line 13: newline at the end of file (checklist)

maxgerman (changes requested):

maxgerman (changes requested):

templates/taxi/driver_form.html
SrMusienko: I have a newline at the end of the file. I don’t know why it is not displayed

maxgerman (approved):
Good job!

PR #673, opened 2024-07-30
Solution

mykola-tarasenko:


taxi/views.py
```diff
@@ -58,13 +60,29 @@ class CarListView(LoginRequiredMixin, generic.ListView):
     queryset = Car.objects.all().select_related("manufacturer")
 
 
-class CarDetailView(LoginRequiredMixin, generic.DetailView):
-    model = Car
+@login_required
+def car_detail_view(request: HttpRequest, pk: int) -> HttpResponse:
+    car = Car.objects.prefetch_related("drivers").get(pk=pk)
```
mykola-tarasenko: Is there any better way to optimize SQL-requests?

vasylhnatiuk: it is good, but use CarDetailView instead of function view

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #679, opened 2024-08-08
Solution

Wallian169:
car_assign
car_create_checkboxes
car_delete_me
create_driver
driver_detail
driver_list
driver_update_license
driver_update_license_invalid_data


taxi/views.py
```diff
@@ -1,9 +1,11 @@
 from django.contrib.auth.decorators import login_required
-from django.shortcuts import render
+from django.contrib.auth.mixins import LoginRequiredMixin
+from django.http import HttpResponse, HttpRequest
+from django.shortcuts import render, redirect
 from django.urls import reverse_lazy
 from django.views import generic
-from django.contrib.auth.mixins import LoginRequiredMixin
 
+from .forms import DriverCreationForm, CarForm, DriverLicenseUpdateForm
```
pavlopro: Use full import instead of relative

taxi/views.py
```diff
@@ -28,6 +30,20 @@ def index(request):
     return render(request, "taxi/index.html", context=context)
 
 
+@login_required
```
pavlopro: Do you use it?

pavlopro (changes requested):

taxi/views.py
Wallian169: No, it is reduntant

pavlopro (approved):
GJ!

PR #685, opened 2024-08-12
implemented taxi service user form and validation

MasakDirt:


MasakDirt:
### **_Driver create page ->_** 

image


### **_License validation ->_** 

image
image


### **_Delete driver page ->_** 

image

### **_Driver list ->_** 

image


### **_Update license page ->_** 

image
image


### **_Car creation page ->_** 

image


### **_Car detail page ->_** 

image
image

taxi/forms.py
```diff
@@ -0,0 +1,57 @@
...
+from django.contrib.auth import get_user_model
+from django.contrib.auth.forms import UserCreationForm
+from django import forms
+from django.core.exceptions import ValidationError
+
+from taxi.models import Driver, Car
+
+
+def validate_license(license_number: str) -> None:
+    if len(license_number) != 8:
```
Oleksl888: you can simplify this by using RegexValidator

taxi/forms.py
```diff
@@ -0,0 +1,57 @@
...
+        model = Driver
+        fields = UserCreationForm.Meta.fields + (
+            "first_name",
+            "last_name",
+            "license_number",
+        )
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    license_number = forms.CharField(
```
Oleksl888: use inheritance here so you don't have to repeat code

taxi/views.py
```diff
@@ -1,9 +1,11 @@
 from django.contrib.auth.decorators import login_required
-from django.shortcuts import render
+from django.contrib.auth.mixins import LoginRequiredMixin
+from django.http import HttpRequest, HttpResponse
+from django.shortcuts import render, redirect, get_object_or_404
 from django.urls import reverse_lazy
 from django.views import generic
-from django.contrib.auth.mixins import LoginRequiredMixin
 
+from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarCreationForm
```

taxi/views.py
```diff
@@ -87,3 +89,32 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+
+
+class DriverDeleteView(LoginRequiredMixin, generic.DeleteView):
+    model = Driver
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+class DriverUpdateView(LoginRequiredMixin, generic.UpdateView):
+    model = Driver
+    form_class = DriverLicenseUpdateForm
```
Oleksl888, line 133: success url here?

taxi/views.py
```diff
@@ -87,3 +89,32 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+
+
+@login_required
```
Oleksl888: You can put this logic into post method of `CarDetailView` instead. The reason is that we don't need to create extra views and url endpoints to handle adding/deleting driver.
The implementation will look something like this:

1. Update form instead on car detail template that will send value 'Add' or 'Delete' to post method to the cardetailview.
2. You override post method in this view and catch those values that you received in if condition and add/remove driver to car through related fields (like you did in methods below).

templates/taxi/car_detail.html
```diff
@@ -3,16 +3,39 @@
...
+       class="btn btn-secondary link-to-page">
       Update
     </a>
   </h1>
   <p>Manufacturer: ({{ car.manufacturer.name }}, {{ car.manufacturer.country }})</p>
-  <h1>Drivers</h1>
+  <h1>
+    Drivers
+    {% if user.is_authenticated %}
+      {% if user not in car.drivers.all %}
```

Oleksl888 (changes requested):

MNDonut (approved):

PR #687, opened 2024-08-17
Implemented solution

dimak20:


dimak20:
image
image
image
image
image
image
image
image
image
image
image

taxi/forms.py
```diff
@@ -0,0 +1,61 @@
...
+    if not re.match(r"^[A-Z]{3}[0-9]{5}$", number):
+        raise ValidationError(
+            "License number must include 5 digits after 3 uppercase letters"
+        )
+
+
+class DriverCreationForm(UserCreationForm):
+    license_number = forms.CharField(
+        required=True,
+        max_length=8,
```
Oleksl888, line 12: can you just use RegexValidator, and MaxLengthValidator to check everything in one place without extra functions?

taxi/forms.py
```diff
@@ -0,0 +1,61 @@
...
+        ]
+    )
+
+    class Meta(UserCreationForm.Meta):
+        model = Driver
+        fields = UserCreationForm.Meta.fields + ("license_number", )
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    license_number = forms.CharField(
```
Oleksl888: why duplicate code here if you can use inheritance instead?

taxi/views.py
```diff
@@ -87,3 +97,34 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+        return self.object.get_absolute_url()
+
+
+class DriverDeleteView(LoginRequiredMixin, generic.DeleteView):
+    model = Driver
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+class AddOrRemoveCar(LoginRequiredMixin, View):
+    def post(self, request, pk, action):
```
Oleksl888: Try to include this logic in CarDetailView - override post method there to add/remove driver to car through related fields. You will also need to adjust your form on a template to pass car_id

taxi/views.py
```diff
@@ -60,17 +61,26 @@ class CarListView(LoginRequiredMixin, generic.ListView):
...
     model = Car
-    fields = "__all__"
+    form_class = CarForm
     success_url = reverse_lazy("taxi:car-list")
 
 
 class CarUpdateView(LoginRequiredMixin, generic.UpdateView):
     model = Car
-    fields = "__all__"
+    form_class = CarForm
```
Oleksl888, line 86: There is no requirement to update form on CarUpdate as per task requirements

templates/taxi/car_detail.html
```diff
@@ -12,7 +12,19 @@ <h1>
     </a>
   </h1>
   <p>Manufacturer: ({{ car.manufacturer.name }}, {{ car.manufacturer.country }})</p>
-  <h1>Drivers</h1>
+  <h1>Drivers
+    {% if has_car %}
+      <form action="{% url 'taxi:add-or-remove-car' pk=car.id action='remove' %}" method="post" style="display:inline;">
+        {% csrf_token %}
+        <input type="submit" value="Delete me from this car" class="btn btn-danger" style="float:right">
+      </form>
```
Oleksl888: To implement this feature thorugh post method of CarDetailView you can keep one form with different inputs (yes/no) that you will later check in post method in your view. And you can include a hidden field here with car_id so you can pull it out in post method and use to create links between car and driver

templates/taxi/driver_form.html
```diff
@@ -0,0 +1,14 @@
+{% extends "base.html" %}
+{% load crispy_forms_filters %}
+{% block content %}
+  <h1>{{ object|yesno:"Update,Create" }} Driver</h1>
+  <form action="" method="post" novalidate>
+    {% csrf_token %}
+    {% if objects.pk %}
+      {% form.license_number|crispy }}
```
Oleksl888: you pass different forms in different views, so you don't need this if, and dont need to pull out `form.license_number` explicitly since DriverLicenseUpdateForm holds only license number field

Oleksl888 (changes requested):

templates/taxi/car_detail.html
dimak20: No need to send hidden car_id since there is a self.get_object() method in detail view

taxi/forms.py
```diff
@@ -0,0 +1,40 @@
...
+            )
+        ]
+    )
+
+    class Meta:
+        model = Driver
+        fields = ("license_number",)
+
+
+class DriverCreationForm(DriverLicenseUpdateForm, UserCreationForm):
```
Oleksl888, line 25: Or you could do something like this:
```
class DriverCreationForm(DriverLicenseUpdateForm): 
    class Meta:
            model = DriverLicenseUpdateForm.Meta.model
            fields = ("username", "license_number", "password")

templates/taxi/car_detail.html
Oleksl888: Good thinking - I overlooked this possibility and suggested a :crutch:

Oleksl888 (approved):
Well done!

PR #697, opened 2024-08-30
"Solution"

OleksiukStepan:
Screenshot 2024-08-30 at 14 55 28
Screenshot 2024-08-30 at 14 56 00
Screenshot 2024-08-30 at 14 56 08
Screenshot 2024-08-30 at 14 58 30
Screenshot 2024-08-30 at 14 58 44
Screenshot 2024-08-30 at 14 58 55
Screenshot 2024-08-30 at 14 59 25
Screenshot 2024-08-30 at 15 00 20
Screenshot 2024-08-30 at 15 00 36
Screenshot 2024-08-30 at 15 04 33
Screenshot 2024-08-30 at 15 05 18
Screenshot 2024-08-30 at 15 05 25
Screenshot 2024-08-30 at 16 20 51

### DIFF
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -68,7 +68,7 @@ class DriverLicenseUpdateForm(DriverCreateForm):
 
     def save(self, commit=True):
         instance = self.instance
-        instance.license_number = self.cleaned_data['license_number']
+        instance.license_number = self.cleaned_data["license_number"]
 
         if commit:
             instance.save()
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -1,9 +1,28 @@
+import re
+
 from django import forms
-from django.contrib.auth.forms import UserCreationForm, UserChangeForm
+from django.contrib.auth.forms import UserCreationForm
 from django.core.exceptions import ValidationError
-from django.core.validators import MaxLengthValidator, RegexValidator
 
 from taxi.models import Driver, Car
+
+
+def validate_license_number(value):
+    if len(value) != 8:
+        raise ValidationError(
+            "License number must be exactly 8 characters long.",
+            code="invalid_length"
+        )
+    if not re.match(r"^[A-Z]{3}", value):
+        raise ValidationError(
+            "The first 3 characters must be uppercase letters",
+            code="invalid_first_three_characters"
+        )
+    if not re.search(r"\d{5}$", value):
+        raise ValidationError(
+            f"The last 5 characters must be numbers {value}",
+            code="invalid_last_five_characters"
+        )
 
 
 class CarCreateForm(forms.ModelForm):
@@ -19,25 +38,8 @@ class CarCreateForm(forms.ModelForm):
 
 
 class DriverCreateForm(UserCreationForm):
-    cars = forms.ModelMultipleChoiceField(
-        queryset=Car.objects.all(),
-        widget=forms.CheckboxSelectMultiple(),
-        required=False,
-    )
     license_number = forms.CharField(
-        validators=[
-            MaxLengthValidator(8),
-            RegexValidator(
-                regex=r"^[A-Z]{3}",
-                message="The first 3 characters must be uppercase letters",
-                code="invalid_first_three_characters"
-            ),
-            RegexValidator(
-                regex=r"\d{5}$",
-                message="The last 5 characters must be numbers",
-                code="invalid_last_five_characters"
-            )
-        ]
+        validators=[validate_license_number],
     )
 
     class Meta(UserCreationForm.Meta):
@@ -46,30 +48,12 @@ class DriverCreateForm(UserCreationForm):
             "username", "first_name", "last_name", "email", "license_number"
         )
 
-    def clean_license_number(self) -> str:
-        license_number = self.cleaned_data["license_number"]
-        if Driver.objects.filter(license_number=license_number).exists():
-            raise ValidationError(
-                "A driver with this license number already exists."
-            )
-        return license_number
 
-
-class DriverLicenseUpdateForm(DriverCreateForm):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        for field_name in list(self.fields.keys()):
-            if field_name not in self.Meta.fields:
-                del self.fields[field_name]
+class DriverLicenseUpdateForm(forms.ModelForm):
+    license_number = forms.CharField(
+        validators=[validate_license_number],
+    )
 
     class Meta:
         model = Driver
         fields = ("license_number",)
-
-    def save(self, commit=True):
-        instance = self.instance
-        instance.license_number = self.cleaned_data["license_number"]
-
-        if commit:
-            instance.save()
-        return instance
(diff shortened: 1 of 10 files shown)


taxi/forms.py
```diff
@@ -0,0 +1,71 @@
...
+    def clean_license_number(self) -> str:
+        license_number = self.cleaned_data["license_number"]
+        if Driver.objects.filter(license_number=license_number).exists():
+            raise ValidationError(
+                "A driver with this license number already exists."
+            )
+        return license_number
+
+
+class DriverCreateForm(DriverForm, UserCreationForm):
```
Oleksl888: this is a bit overcomplicated - you can make DriverCreateForm with set of fields and validators and then you can inherit UpdateForm from it and only override fields in Meta class

taxi/views.py
```diff
@@ -1,9 +1,10 @@
 from django.contrib.auth.decorators import login_required
-from django.shortcuts import render
+from django.shortcuts import render, get_object_or_404, redirect
 from django.urls import reverse_lazy
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
 
+from .forms import DriverCreateForm, DriverLicenseUpdateForm, CarCreateForm
```

taxi/views.py
```diff
@@ -28,6 +29,16 @@ def index(request):
     return render(request, "taxi/index.html", context=context)
 
 
+@login_required
```

templates/taxi/driver_form.html
```diff
@@ -0,0 +1,16 @@
+{% extends "base.html" %}
+{% load crispy_forms_filters %}
+
+{% block content %}
+  {% if form_type == "license" %}
```
Oleksl888: why do you need this if? you can simply update `yesno:"Update..`

templates/taxi/car_detail.html
```diff
@@ -3,20 +3,37 @@
...
-    <a href="{% url 'taxi:car-update' pk=car.id %}" class="btn btn-secondary link-to-page">
-      Update
-    </a>
   </h1>
   <p>Manufacturer: ({{ car.manufacturer.name }}, {{ car.manufacturer.country }})</p>
-  <h1>Drivers</h1>
+  <h1>
+    Drivers
+    {% if user.is_authenticated %}
+      <form action="{% url 'taxi:toggle-driver' car.id %}" method="post">
```
Oleksl888: Adjust the form so it can set driver in CarDetailView

taxi/views.py
```diff
@@ -87,3 +104,26 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    form_class = DriverCreateForm
+
+
+class DriverLicenseUpdateView(LoginRequiredMixin, generic.UpdateView):
+    model = Driver
+    success_url = reverse_lazy("taxi:driver-list")
+    template_name = "taxi/driver_form.html"
+    form_class = DriverLicenseUpdateForm
+
+    def get_context_data(self, **kwargs):
```
Oleksl888: you don't need to pass extra variables in context just to display a different h1 tag

Oleksl888 (changes requested):

templates/taxi/driver_form.html
OleksiukStepan: The previous version of my code included a driver info update feature, for which it was logical. 
I updated this block for the current version.

taxi/forms.py
```diff
@@ -0,0 +1,75 @@
...
+        required=False,
+    )
+
+    class Meta:
+        model = Car
+        fields = "__all__"
+
+
+class DriverCreateForm(UserCreationForm):
+    cars = forms.ModelMultipleChoiceField(
```
Oleksl888: you don't set the cars on driver creation so this is unnecessary

taxi/forms.py
```diff
@@ -0,0 +1,75 @@
...
+        ]
+    )
+
+    class Meta(UserCreationForm.Meta):
+        model = Driver
+        fields = (
+            "username", "first_name", "last_name", "email", "license_number"
+        )
+
+    def clean_license_number(self) -> str:
```
Oleksl888: we have a unique constraint on a model for this

taxi/forms.py
```diff
@@ -0,0 +1,75 @@
...
+        license_number = self.cleaned_data["license_number"]
+        if Driver.objects.filter(license_number=license_number).exists():
+            raise ValidationError(
+                "A driver with this license number already exists."
+            )
+        return license_number
+
+
+class DriverLicenseUpdateForm(DriverCreateForm):
+    def __init__(self, *args, **kwargs):
```
Oleksl888: no need to override init - overriding fields in Meta class would be enough

taxi/forms.py
```diff
@@ -0,0 +1,75 @@
...
+            if field_name not in self.Meta.fields:
+                del self.fields[field_name]
+
+    class Meta:
+        model = Driver
+        fields = ("license_number",)
+
+    def save(self, commit=True):
+        instance = self.instance
+        instance.license_number = self.cleaned_data["license_number"]
```
Oleksl888: No need to override this method - license number will be updated anyway

taxi/views.py
```diff
@@ -28,6 +29,16 @@ def index(request):
     return render(request, "taxi/index.html", context=context)
 
 
+@login_required
+def toggle_driver(request, car_id):
```
Oleksl888: you don't need this method anymore since you added this logic in CarDetailView

templates/taxi/driver_detail.html
```diff
@@ -23,4 +23,12 @@ <h4>Cars</h4>
       <p>No cars!</p>
     {% endfor %}
   </div>
+  <div class="button-container">
+    <a style="float: left" href="{% url 'taxi:driver-update' pk=driver.id %}" class="btn btn-secondary link-to-page">
+      Update License number
```
Oleksl888, line 28: Is it correct label for this operation?

Oleksl888 (changes requested):

taxi/forms.py
OleksiukStepan: Without init and save, it doesn't work correctly. I have changed the logic by separating the validation into its own function.

taxi/forms.py
```diff
@@ -0,0 +1,59 @@
...
+    )
+
+    class Meta(UserCreationForm.Meta):
+        model = Driver
+        fields = (
+            "username", "first_name", "last_name", "email", "license_number"
+        )
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
```
Oleksl888, line 52: So you saying this will not work for you?
```
class DriverLicenseUpdateForm(DriverCreateForm):

    class Meta:
        fields = ("license_number",)

Oleksl888 (approved):
Well done!

taxi/forms.py
OleksiukStepan: No, it doesn’t( 
Screenshot 2024-09-24 at 15 47 02
Screenshot 2024-09-24 at 15 45 35
I’ve tried it again, and it disrupts tests and the logic for updating the license number on the site page.

PR #698, opened 2024-08-30
Solution

DSahalatyi:
**Create driver page**
image

**Delete driver page**
image

**Driver license update page with validation**
image

**Driver license validation on driver creation form**
image

**Update and delete buttons on driver page**
image

**Checkboxes for drivers selection on car creation page**
image

**Add me to this car button on car page**
image

**Delete me from this car button on car page**
image

### DIFF
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -1,12 +1,19 @@
 from django import forms
 from django.contrib.auth import get_user_model
-from django.contrib.auth.forms import UserCreationForm
+from django.contrib.auth.forms import UserCreationForm, UserChangeForm
+from django.core.validators import RegexValidator
 from django.forms import CheckboxSelectMultiple
 
 from taxi.models import Driver, Car
 
 
 class DriverCreationForm(UserCreationForm):
+    license_number = forms.CharField(
+        validators=[
+            RegexValidator(r"^[A-Z]{3}\d{5}$"),
+        ],
+    )
+
     class Meta(UserCreationForm):
         model = Driver
         fields = (
@@ -20,6 +27,12 @@ class DriverCreationForm(UserCreationForm):
 
 
 class DriverLicenseUpdateForm(forms.ModelForm):
+    license_number = forms.CharField(
+        validators=[
+            RegexValidator(r"^[A-Z]{3}\d{5}$"),
+        ],
+    )
+
     class Meta:
         model = Driver
         fields = ("license_number",)
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -1,11 +1,6 @@
 from django import forms
 from django.contrib.auth import get_user_model
 from django.contrib.auth.forms import UserCreationForm
-from django.core.validators import (
-    MaxLengthValidator,
-    MinLengthValidator,
-    RegexValidator,
-)
 from django.forms import CheckboxSelectMultiple
 
 from taxi.models import Driver, Car
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -1,0 +1,42 @@
+from django import forms
+from django.contrib.auth import get_user_model
+from django.contrib.auth.forms import UserCreationForm
+from django.core.validators import (
+    MaxLengthValidator,
+    MinLengthValidator,
+    RegexValidator,
+)
+from django.forms import CheckboxSelectMultiple
+
+from taxi.models import Driver, Car
+
+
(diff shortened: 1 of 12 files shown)


taxi/forms.py
```diff
@@ -0,0 +1,42 @@
+from django import forms
+from django.contrib.auth import get_user_model
+from django.contrib.auth.forms import UserCreationForm
+from django.core.validators import (
+    MaxLengthValidator,
+    MinLengthValidator,
+    RegexValidator,
+)
```
MNDonut: do you use these validators here?

taxi/models.py
```diff
@@ -15,7 +20,15 @@ def __str__(self):
...
 
 class Driver(AbstractUser):
-    license_number = models.CharField(max_length=255, unique=True)
+    license_number = models.CharField(
+        max_length=255,
+        unique=True,
+        validators=[
+            MinLengthValidator(8),
+            MaxLengthValidator(8),
+            RegexValidator(r"^[A-Z]{3}\d{5}$"),
```
MNDonut: do you need to have here min and max validators if you regex has ^ and & with 8 characters between them?

templates/taxi/car_detail.html
```diff
@@ -12,7 +12,20 @@ <h1>
     </a>
   </h1>
   <p>Manufacturer: ({{ car.manufacturer.name }}, {{ car.manufacturer.country }})</p>
-  <h1>Drivers</h1>
+  <div class="driver-heading">
+    <h1>Drivers</h1>
+    {% if user in car.drivers.all %}
+      <form action="{% url "taxi:delete-current-driver" pk=car.id %}" method="post">
+        {% csrf_token %}
```

MNDonut (changes requested):

taxi/models.py
```diff
@@ -15,7 +18,13 @@ def __str__(self):
 
 
 class Driver(AbstractUser):
-    license_number = models.CharField(max_length=255, unique=True)
+    license_number = models.CharField(
```
Oleksl888, line 18: you should add validation on a form in this exerices

taxi/views.py
```diff
@@ -1,9 +1,12 @@
 from django.contrib.auth.decorators import login_required
+from django.core.exceptions import ValidationError
 from django.shortcuts import render
 from django.urls import reverse_lazy
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
+from django.http import HttpResponseRedirect
 
+from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm
```

taxi/views.py
```diff
@@ -87,3 +95,46 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    success_url = reverse_lazy("taxi:car-list")
+
+
+class DriverLicenseUpdateView(LoginRequiredMixin, generic.UpdateView):
+    model = Driver
+    form_class = DriverLicenseUpdateForm
+    template_name = "taxi/driver_license_update.html"
+
+
+def add_current_driver(request, pk):
```

templates/taxi/car_detail.html
```diff
@@ -12,7 +12,20 @@ <h1>
...
+        <input class="btn btn-danger" type="submit" value="Delete me from this car"/>
+      </form>
+    {% else %}
```
Oleksl888: Keep only one form and adjust it so it can set driver in CarDetailView

Oleksl888 (changes requested):

taxi/views.py
```diff
@@ -64,10 +64,28 @@ class CarListView(LoginRequiredMixin, generic.ListView):
...
+        car = Car.objects.get(id=kwargs["pk"])
+        if action == "add" and self.request.user not in car.drivers.all():
+            car.drivers.add(self.request.user)
+            car.save()
+            return HttpResponseRedirect(
+                reverse_lazy("taxi:car-detail", kwargs={"pk": car.pk})
+            )
+
+        if action == "delete" and self.request.user in car.drivers.all():
+            car.drivers.remove(self.request.user)
```
MNDonut: 1. use .exists() instead of "in" operator
2. no need to call .save() method after .add and .remove
3. you have a common part in both if statements, so don't duplicate the code

MNDonut (changes requested):

MNDonut (approved):
what a beauty!

PR #707, opened 2024-09-08
solution

BastovOleksandr:
car create page
car detail page 1
car detail page 2
driver create button
driver create page
driver delete page
driver update   delete buttons
driver update license invalid
driver update license page


taxi/urls.py
```diff
@@ -42,10 +47,33 @@
...
+    path(
+        "car/<int:pk>/assign-driver/",
+        car_assign_driver,
+        name="car-assign-driver"
+    ),
+    path(
+        "car/<int:pk>/remove-driver/",
+        car_remove_driver,
+        name="car-remove-driver"
+    ),
```
MNDonut: cars*

templates/taxi/car_detail.html
```diff
@@ -1,22 +1,30 @@
...
-    {{ car.model }}
-    <a href="{% url 'taxi:car-delete' pk=car.id %}" class="btn btn-danger link-to-page">
-      Delete
-    </a>
-
-    <a href="{% url 'taxi:car-update' pk=car.id %}" class="btn btn-secondary link-to-page">
-      Update
-    </a>
+    Drivers
+    {% if user in car.drivers.all %}
```
MNDonut, line 11: you use here backward relationship but did you fetched it from db? because otherwise here will be N+1 issue

templates/taxi/car_detail.html
```diff
@@ -1,22 +1,30 @@
...
+      <a href="{% url 'taxi:car-assign-driver' pk=car.pk %}" class="btn btn-success link-to-page">
+        Assign me to this car
+      </a>
+    {% endif %}
   </h1>
-  <p>Manufacturer: ({{ car.manufacturer.name }}, {{ car.manufacturer.country }})</p>
-  <h1>Drivers</h1>
   <hr>
   <ul>
     {% for driver in car.drivers.all %}
```
MNDonut, line 23: you use here many to many relationship but did you fetched it from db? because otherwise here will be N+1 issue

MNDonut (changes requested):

taxi/forms.py
```diff
@@ -0,0 +1,79 @@
...
+        required=True,
+        validators=[
+            RegexValidator(
+                LICENSE_NUMBER_PATTERN,
+                message=LICENSE_NUMBER_MESSAGE
+            ),
+        ],
+        help_text=LICENSE_NUMBER_HELP_TEXT
+    )
+    cars = forms.ModelMultipleChoiceField(
```
Oleksl888: are you sure that m2m relations will be recorded using generic CreateView?

taxi/forms.py
```diff
@@ -0,0 +1,79 @@
...
+            "first_name",
+            "last_name",
+            "email",
+            "license_number",
+            "cars"
+        )
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    license_number = forms.CharField(
```
Oleksl888, line 28: use inheritance to avoid code repetition

taxi/forms.py
```diff
@@ -0,0 +1,79 @@
...
+        widget=forms.CheckboxSelectMultiple,
+        required=False
+    )
+
+    class Meta:
+        model = Car
+        fields = ("model", "manufacturer", "drivers",)
+
+
+class CarUpdateForm(forms.ModelForm):
```
Oleksl888: this form is exactly the same as above + there is no requirement to override m2m fields on updateview

taxi/views.py
```diff
@@ -1,9 +1,15 @@
+from django.contrib import messages
 from django.contrib.auth.decorators import login_required
-from django.shortcuts import render
+from django.shortcuts import render, redirect, get_object_or_404
 from django.urls import reverse_lazy
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
 
+from .forms import (
```

taxi/views.py
```diff
@@ -60,25 +66,48 @@ class CarListView(LoginRequiredMixin, generic.ListView):
...
 
 
 class CarUpdateView(LoginRequiredMixin, generic.UpdateView):
     model = Car
-    fields = "__all__"
+    form_class = CarUpdateForm
     success_url = reverse_lazy("taxi:car-list")
 
 
+class CarUpdateDriverView(LoginRequiredMixin, generic.UpdateView):
```
Oleksl888: you should have only one CarUpdateView

taxi/views.py
```diff
@@ -60,25 +66,48 @@ class CarListView(LoginRequiredMixin, generic.ListView):
...
+    model = Car
+    success_url = reverse_lazy("taxi:car-detail")
+
+
 class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
     model = Car
     success_url = reverse_lazy("taxi:car-list")
 
 
+@login_required
```
Oleksl888, line 105: You can put this logic into post method of CarDetailView. This way you don't need to create extra views and url endpoints to handle adding/deleting driver.

templates/taxi/car_detail.html
```diff
@@ -1,22 +1,30 @@
...
-    <a href="{% url 'taxi:car-delete' pk=car.id %}" class="btn btn-danger link-to-page">
-      Delete
-    </a>
-
-    <a href="{% url 'taxi:car-update' pk=car.id %}" class="btn btn-secondary link-to-page">
-      Update
-    </a>
+    Drivers
+    {% if user in car.drivers.all %}
+      <a href="{% url 'taxi:car-remove-driver' pk=car.pk %}" class="btn btn-danger link-to-page">
```
Oleksl888: Use form here instead so it can set driver in CarDetailView

templates/taxi/driver_license_update.html
Oleksl888, line 1: there can be one template for create/update driver

Oleksl888 (changes requested):

MNDonut (approved):

PR #709, opened 2024-09-13
implement py-taxi-service-form-and-validation

ValeriiaKyr:
Снимок экрана 2024-09-13 в 16 05 26
Снимок экрана 2024-09-13 в 16 06 10
Снимок экрана 2024-09-13 в 16 06 25
Снимок экрана 2024-09-13 в 16 06 37

### DIFF
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -1,0 +1,47 @@
+from django import forms
+from django.contrib.auth import get_user_model
+from django.contrib.auth.forms import UserCreationForm
+from django.core.exceptions import ValidationError
+
+from taxi.models import Driver, Car
+
+
+class CarCreateForm(forms.ModelForm):
+    drivers = forms.ModelMultipleChoiceField(
+        queryset=get_user_model().objects.all(),
+        widget=forms.CheckboxSelectMultiple,
+        required=False,
+    )
+
+    class Meta:
+        model = Car
+        fields = "__all__"
+
+
+class DriverCreationForm(UserCreationForm):
+    class Meta(UserCreationForm.Meta):
+        model = Driver
+        fields = UserCreationForm.Meta.fields + (
+            "first_name",
+            "last_name",
+            "license_number",
+        )
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+
+    class Meta:
+        model = Driver
+        fields = ("license_number",)
+
+    def clean_license_number(self):
+        license_number = self.cleaned_data["license_number"]
+        part1 = license_number[:3]
+        part2 = license_number[3:]
+        if len(license_number) != 8:
+            raise ValidationError("Consist only of 8 characters")
+        elif not str(part1).isalpha() or not str(part1).isupper():
+            raise ValidationError("First 3 characters are uppercase letters")
+        elif not part2.isdigit():
+            raise ValidationError("Last 5 characters are digits")
+        return license_number
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -18,22 +18,7 @@ class CarCreateForm(forms.ModelForm):
         fields = "__all__"
 
 
-class DriverCreationForm(UserCreationForm):
-    class Meta(UserCreationForm.Meta):
-        model = Driver
-        fields = UserCreationForm.Meta.fields + (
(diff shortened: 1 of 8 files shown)


taxi/urls.py
```diff
@@ -43,9 +47,26 @@
...
     path("cars/<int:pk>/delete/", CarDeleteView.as_view(), name="car-delete"),
     path("drivers/", DriverListView.as_view(), name="driver-list"),
+    path("drivers/create/", DriverCreateView.as_view(), name="driver-create"),
+    path(
+        "drivers/<int:pk>/delete/",
+        DriverDeleteView.as_view(),
+        name="driver-delete"
+    ),
+    path(
+        "drivers/<int:pk>/updete/",
```
MNDonut: update/*

taxi/views.py
```diff
@@ -87,3 +90,31 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    model = Driver
+    form_class = DriverLicenseUpdateForm
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+@login_required()
+def assign_to_car(request, pk):
+    driver = Driver.objects.get(id=request.user.id)
+    car = Car.objects.get(id=pk)
+    if car in driver.cars.all():
```
MNDonut: use .exists() so that db will check it using EXISTS and WHERE, because in this way you make query to db, it fetches all records and the python loops through all objects to check if car is in this list

templates/taxi/car_detail.html
```diff
@@ -12,6 +12,11 @@ <h1>
     </a>
   </h1>
   <p>Manufacturer: ({{ car.manufacturer.name }}, {{ car.manufacturer.country }})</p>
+  {% if car in user.cars.all %}
+      <a href="{% url 'taxi:car-assign_to_car' pk=car.id %}" class="btn btn-danger link-to-page">Delete me from this car</a>
+    {% else %}
+      <a href="{% url 'taxi:car-assign_to_car' pk=car.id %}" class="btn btn-secondary link-to-page">Assign me to this car</a>
```

MNDonut (changes requested):

taxi/forms.py
```diff
@@ -0,0 +1,47 @@
...
+        )
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+
+    class Meta:
+        model = Driver
+        fields = ("license_number",)
+
+    def clean_license_number(self):
```
Oleksl888: license number must be validated on Driver Create as well (you should use inheritance to avoid code repetition)

taxi/views.py
```diff
@@ -87,3 +92,31 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+
+
+class DriverUpdateView(LoginRequiredMixin, generic.UpdateView):
+    model = Driver
+    form_class = DriverLicenseUpdateForm
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+@login_required()
+def assign_to_car(request, pk):
```
Oleksl888: You can put this logic into post method of CarDetailView. This way you don't need to create extra view and url endpoint to handle adding/deleting driver.
*You will need to create a form on car detail template to send value to the CarDetailView + clean up urls.py

taxi/views.py
```diff
@@ -1,9 +1,13 @@
...
 from django.contrib.auth.decorators import login_required
-from django.shortcuts import render
+from django.shortcuts import render, redirect
 from django.urls import reverse_lazy
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
+from django.http import HttpRequest, HttpResponse, HttpResponseRedirect
+from django.shortcuts import get_object_or_404
 
+from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarCreateForm
```
Oleksl888: Change relative import here

templates/taxi/car_detail.html
Oleksl888: Use a form instead of hyperlinks to send value to CarDetailView

Oleksl888 (changes requested):

Arsen-hrynevych (approved):
Overall ok

PR #710, opened 2024-09-13
Solution

olena4reunited:
Screenshot from 2024-09-13 18-28-53
Screenshot from 2024-09-13 18-29-06
Screenshot from 2024-09-13 18-29-24
Screenshot from 2024-09-13 18-29-34
Screenshot from 2024-09-13 18-29-40
Screenshot from 2024-09-13 18-29-55


taxi/views.py
```diff
@@ -61,17 +63,40 @@ class CarListView(LoginRequiredMixin, generic.ListView):
...
+
+
+@login_required
+def toggle_driver(request: HttpRequest, pk: int) -> HttpResponse:
+    car = get_object_or_404(Car, pk=pk)
+
+    if request.user in car.drivers.all():
+        car.drivers.remove(request.user)
+    else:
+        car.drivers.add(request.user)
```
MNDonut: 1. you don't prefetch drivers
2. it's better to use .exists() method because in this case you fetch 1 000 000 records and then python loops through all objects to find if there is request.user. in case you use .exists(), db will check it using "where" condition

taxi/views.py
```diff
@@ -61,17 +63,40 @@ class CarListView(LoginRequiredMixin, generic.ListView):
 class CarDetailView(LoginRequiredMixin, generic.DetailView):
     model = Car
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        car = self.get_object()
+        context["driver_ids"] = list(car.drivers.values_list("id", flat=True))
```
MNDonut: drivers aren't prefetched

MNDonut (changes requested):

taxi/models.py
```diff
@@ -15,7 +16,17 @@ def __str__(self):
 
 
 class Driver(AbstractUser):
-    license_number = models.CharField(max_length=255, unique=True)
+    license_number = models.CharField(
```
Oleksl888: validation should happen on the form in this task

taxi/views.py
```diff
@@ -1,9 +1,13 @@
...
 from django.contrib.auth.decorators import login_required
-from django.shortcuts import render
+from django.db.models import Prefetch
+from django.http import HttpRequest, HttpResponse
+from django.shortcuts import render, get_object_or_404, redirect
 from django.urls import reverse_lazy
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
 
+from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm
```

taxi/views.py
```diff
@@ -61,17 +65,55 @@ class CarListView(LoginRequiredMixin, generic.ListView):
...
+                        "first_name",
+                        "last_name"
+                    ),
+                )
+            )
+        )
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["drivers"] = self.object.drivers.all()
```
Oleksl888: do prefetch here - then you won't to override method above

taxi/views.py
```diff
@@ -61,17 +65,55 @@ class CarListView(LoginRequiredMixin, generic.ListView):
...
+
+        return context
+
+
+@login_required
```
Oleksl888: Include this logic into post method of CarDetailView. This way you don't need to create extra view and url endpoint to handle adding/deleting driver.

taxi/views.py
```diff
@@ -84,6 +126,28 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
     paginate_by = 5
 
 
+class DriverCreateView(LoginRequiredMixin, generic.CreateView):
+    model = Driver
+    form_class = DriverCreationForm
+
+    def get_success_url(self):
```
Oleksl888: Here you don't need to set get_success_url because get_absolute_url is defined in your model - this should redirect to DriverDetail under the hood

taxi/views.py
```diff
@@ -84,6 +126,28 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    model = Driver
+    form_class = DriverLicenseUpdateForm
+
+    def get_success_url(self):
+        return self.object.get_absolute_url()
+
+
+class DriverDeleteView(LoginRequiredMixin, generic.DeleteView):
+    model = Driver
+    success_url = reverse_lazy("login")
```
Oleksl888: why would it redirect to login page? this is not logout

templates/taxi/car_detail.html
```diff
@@ -12,10 +12,26 @@ <h1>
     </a>
   </h1>
   <p>Manufacturer: ({{ car.manufacturer.name }}, {{ car.manufacturer.country }})</p>
-  <h1>Drivers</h1>
+    <h1>
+      Drivers
+      {% if user.is_authenticated %}
+      <form action="{% url 'taxi:toggle-driver' pk=car.id %}" method="post">
```
Oleksl888: adjust this form so it can set driver in CarDetailView

templates/taxi/driver_form.html
```diff
@@ -0,0 +1,12 @@
+{% extends "base.html" %}
+{% load crispy_forms_filters %}
+
+{% block content %}
+  <h1>Create driver</h1>
```
Oleksl888: you should add conditionals here for update driver as well

Oleksl888 (changes requested):

taxi/models.py
```diff
@@ -1,3 +1,4 @@
+from django.core.validators import RegexValidator
```
Oleksl888: remove unused imports

taxi/forms.py
```diff
@@ -0,0 +1,72 @@
...
+
+        return license_number
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    class Meta:
+        model = Driver
+        fields = ("license_number",)
+
+    def clean_license_number(self):
```
Oleksl888, line 49: if you inherit this form from  DriverCreationForm you will not need to duplicate code here

Oleksl888 (changes requested):

MNDonut (changes requested):
You didn't push any new commits ;)

MNDonut (approved):

PR #712, opened 2024-09-14
Solution Politov

Alexandr-Politov:


Alexandr-Politov:
image
image
image
image
image
image

taxi/forms.py
```diff
@@ -0,0 +1,51 @@
...
+from django.contrib.admin.widgets import FilteredSelectMultiple
+from django.core.validators import RegexValidator
+from django.contrib.auth.forms import UserCreationForm
+
+from taxi.models import Driver, Car
+
+
+class DriverForm(UserCreationForm):
+    license_number = forms.CharField(
+        max_length=8,
```
Oleksl888, line 11: why not add max length validator here and show beautiful message if validation will not go through?

taxi/forms.py
```diff
@@ -0,0 +1,51 @@
...
+
+    class Meta(UserCreationForm.Meta):
+        model = Driver
+        fields = UserCreationForm.Meta.fields + (
+            "first_name", "last_name", "license_number"
+        )
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    license_number = forms.CharField(
```
Oleksl888: you can play with inheritance here and avoid code repetition

taxi/forms.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+
+    class Meta:
+        model = Driver
+        fields = ["license_number"]
+
+
+class CarForm(forms.ModelForm):
+    drivers = forms.ModelMultipleChoiceField(
+        queryset=Driver.objects.prefetch_related("cars"),
```
Oleksl888: why do you need to prefetch cars here?

taxi/views.py
```diff
@@ -64,13 +67,13 @@ class CarDetailView(LoginRequiredMixin, generic.DetailView):
...
     model = Car
-    fields = "__all__"
+    form_class = CarForm
     success_url = reverse_lazy("taxi:car-list")
 
 
 class CarUpdateView(LoginRequiredMixin, generic.UpdateView):
     model = Car
-    fields = "__all__"
+    form_class = CarForm
```
Oleksl888, line 90: > On car creation page switch to checkboxes widget for assigning drivers to car.

It looks like you need to use this custom form only on Create as per task requirements

taxi/views.py
```diff
@@ -87,3 +104,30 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
     model = Driver
     queryset = Driver.objects.all().prefetch_related("cars__manufacturer")
+
+
+class DriverCreateView(LoginRequiredMixin, generic.CreateView):
+    model = Driver
+    form_class = DriverForm
+
+
+class DriverUpdateView(LoginRequiredMixin, generic.UpdateView):
```
Oleksl888: As per requirements you should keep only one view with Driver Update - that will be one with only driver license update

taxi/views.py
```diff
@@ -87,3 +104,30 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
 class DriverDetailView(LoginRequiredMixin, generic.DetailView):
     model = Driver
     queryset = Driver.objects.all().prefetch_related("cars__manufacturer")
+
+
+class DriverCreateView(LoginRequiredMixin, generic.CreateView):
+    model = Driver
+    form_class = DriverForm
```
Oleksl888, line 111: success url here?

taxi/views.py
```diff
@@ -79,6 +82,20 @@ class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
     success_url = reverse_lazy("taxi:car-list")
 
 
+@login_required
```
Oleksl888: Try to include this logic in CarDetailView - override `post` method to add/remove driver to car through related fields. You will also need to adjust your form on a template to pass car_id

templates/taxi/car_detail.html
```diff
@@ -11,12 +12,21 @@ <h1>
...
   <h1>Drivers</h1>
   <hr>
   <ul>
     {% for driver in car.drivers.all %}
       <li>{{ driver.username }} ({{ driver.first_name }} {{ driver.last_name }})</li>
     {% endfor %}
   </ul>
+  
+  {% if user in car.drivers.all %}
+    <a href="{% url 'taxi:remove-from-car' car.id %}" class="btn btn-warning">Delete me from this car</a>
```
Oleksl888: To implement this feature thorugh post method of CarDetailView you will need to change links to  form with different inputs and different values that you will later compare in your view. Also you can include a hidden field here with car_id so you can pull it out in post method and use to create links between car and driver

templates/taxi/driver_list.html
```diff
@@ -1,30 +1,36 @@
 {% extends "base.html" %}
 
+
+{% block title %}
+	<title>driver list</title>
```
Oleksl888: check indentation

Oleksl888 (changes requested):

taxi/forms.py
Alexandr-Politov: It's always better to prevent user from making mistakes, then to give him messages when mistake is already done.
I have added information message about license number parameters "
image

taxi/forms.py
Alexandr-Politov: when I'm trying to ingerit from DriverForm, password1 and password2 firelds appear and so I can't pass the tests.
class DriverLicenseUpdateForm(DriverForm):
    class Meta(DriverForm.Meta):
        fields = ["license_number"]

taxi/views.py
Alexandr-Politov: Isn't it better to do alitle more then just requirements say?

taxi/views.py
Alexandr-Politov: django redirects user to this new created user page by default.

taxi/views.py
Alexandr-Politov: what is "this logic" in this case? I can't understand what I should do.

taxi/forms.py
Oleksl888: actually max_length, should trigger MaxLengthValidator under the hood anyway so yeah it's not really critical here

taxi/forms.py
Oleksl888: That's probably because you inherit DriverForm from UserCreationForm - use forms.ModelForm on parent class and set the fields explicitly like so : `("username",  "password","license_number",)`. Then you can inherit update form from DriverForm and reset Meta class only

taxi/views.py
Oleksl888: In this task the exercise is about correctly implementing overriding of form on UpdateView. And what is the point of having a full update and also update of one field?

taxi/views.py
Oleksl888: You can put logic from assign_to_car and  remove_from_car into post method of CarDetailView. The reason is that we don't need to create two extra views and two extra endpoints to handle adding/deleting driver. 
1. Add an empty form on car detail template that will send value 'Add' or 'Delete'
3. You override post method in this view and check values that you recieved in if condition. 
4. You get user from `request.user` and car_id you can pass from template(make a separate hidden field) or you can take car object from `self.get_object()` in the view.
5. Assign/unassign driver to the car

Oleksl888 (changes requested):

taxi/forms.py
Alexandr-Politov: all the same this way, if I'm setting the "password" field to DriverForm, it appears in license update form as well: 
image
image

taxi/forms.py
Alexandr-Politov: because Driver and Cars have many-to-many relations.

taxi/views.py
Alexandr-Politov: OK. I have deleted unnecessary updating form, view and link. Now everything (even updating license) works like it should.

taxi/forms.py
Alexandr-Politov: OK. I have deleted unnecessary updating form, view and link. Now everything (even updating license) works like it should.

taxi/views.py
Alexandr-Politov: image

taxi/forms.py
```diff
@@ -0,0 +1,51 @@
...
+            "first_name", "last_name", "license_number"
+        )
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    license_number = forms.CharField(
+        max_length=8,
+        validators=[
+            RegexValidator(
+                regex=r"^[A-Z]{3}\d{5}$",
```
Oleksl888: validation is still duplicated. The below code should work - if not then lets discuss on QnA:
```
class DriverLicenseUpdateForm(DriverForm):

    class Meta:
        model = DriverForm.Meta.model
        fields = ("license_number",)

File: taxi/urls.py [hunk 11]
```diff
@@ -1,5 +1,6 @@
 from django.urls import path
 
+from .models import Driver
```
Oleksl888: what is this import for?

taxi/views.py
```diff
@@ -1,9 +1,12 @@
 from django.contrib.auth.decorators import login_required
-from django.shortcuts import render
+from django.http import HttpRequest, HttpResponse
+from django.shortcuts import render, get_object_or_404, redirect
 from django.urls import reverse_lazy
+from django.urls.base import reverse
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
 
+from .forms import DriverForm, DriverLicenseUpdateForm, CarForm
```
Oleksl888, line 9: import shouldn't be relative

taxi/views.py
```diff
@@ -61,16 +64,43 @@ class CarListView(LoginRequiredMixin, generic.ListView):
...
+
+        if action == "Add":
+            car.drivers.add(request.user)
+        elif action == "Delete":
+            car.drivers.remove(request.user)
+
+        return redirect(reverse("taxi:car-detail", args=[car.id]))
+
+
+# @login_required
```
Oleksl888: clean up these comments please

taxi/views.py
Oleksl888: It actually redirects because Driver model has `get_absolute_url` method defined :nerd_face:

templates/taxi/car_detail.html
```diff
@@ -11,12 +12,31 @@ <h1>
...
+  
   <h1>Drivers</h1>
   <hr>
   <ul>
     {% for driver in car.drivers.all %}
       <li>{{ driver.username }} ({{ driver.first_name }} {{ driver.last_name }})</li>
     {% endfor %}
   </ul>
+
+  {% if user in car.drivers.all %}
```
Oleksl888: You need to fix N+1 problem on a car detail view here

templates/taxi/car_detail.html
```diff
@@ -11,12 +12,31 @@ <h1>
...
+    <form method="post" action="{% url 'taxi:car-detail' car.id %}">
+      {% csrf_token %}
+      <!-- Hidden field to specify the action -->
+      <input type="hidden" name="action" value="Delete">
+      <button type="submit">Remove me from driver's list</button>
+    </form>
+  {% else %}
+    <form method="post" action="{% url 'taxi:car-detail' car.id %}">
```
Oleksl888: keep only one form but add condition to show only one input at a time

Oleksl888 (changes requested):

MNDonut (approved):

PR #713, opened 2024-09-15
implement py-taxi-service-user-form-and-validation

Maksym-Turenko:
Снимок экрана 2024-09-15 144816
Снимок экрана 2024-09-15 144825

### DIFF
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -11,10 +11,11 @@ class DriverLicenseUpdateForm(forms.ModelForm):
         max_length=8,
         validators=[
             RegexValidator(
-                regex=r'^[A-Z]{3}\d{5}$',
+                regex=r"^[A-Z]{3}\d{5}$",
                 message=(
                     "The license number must consist of 8 characters: "
-                    "the first 3 characters are capital letters, the last 5 characters are numbers."
+                    "the first 3 characters are capital letters, "
+                    "the last 5 characters are numbers."
                 )
             )
         ]
@@ -30,7 +31,13 @@ class DriverCreateForm(forms.ModelForm):
 
     class Meta:
         model = Driver
-        fields = ["username", "first_name", "last_name", "license_number", "password"]
+        fields = [
+            "username",
+            "first_name",
+            "last_name",
+            "license_number",
+            "password"
+        ]
 
     def clean_license_number(self):
         license_number = self.cleaned_data.get("license_number")
@@ -39,7 +46,8 @@ class DriverCreateForm(forms.ModelForm):
         if not re.match(pattern, license_number):
             raise forms.ValidationError(
                 "The driver's license number must consist of 8 characters: "
-                "the first 3 characters are capital letters, the last 5 characters are numbers."
+                "the first 3 characters are capital letters, "
+                "the last 5 characters are numbers."
             )
         return license_number
 
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -1,6 +1,5 @@
 from django import forms
 from .models import Driver
-import re
 
 
 class DriverLicenseUpdateForm(forms.ModelForm):
@@ -9,19 +8,21 @@ class DriverLicenseUpdateForm(forms.ModelForm):
         fields = ["license_number"]
 
     def clean_license_number(self):
-        license_number = self.cleaned_data.get("license_number")
+        license_number = self.cleaned_data["license_number"]
 
         if len(license_number) != 8:
             raise forms.ValidationError(
-                "The license number must contain 8 characters."
-            )
+                "The license number must be 8 characters long.")
 
-        if not re.match(r"^[A-Z]{3}", license_number):
+        if (
+                not license_number[:3].isalpha() or not
+                license_number[:3].isupper()
+        ):
             raise forms.ValidationError(
                 "The first 3 characters must be capital letters."
             )
 
-        if not re.match(r"[0-9]{5}$", license_number):
+        if not license_number[3:].isdigit():
             raise forms.ValidationError(
                 "The last 5 characters must be numbers."
             )
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -1,0 +1,29 @@
+from django import forms
+from .models import Driver
+import re
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    class Meta:
+        model = Driver
+        fields = ["license_number"]
+
+    def clean_license_number(self):
+        license_number = self.cleaned_data.get("license_number")
+
+        if len(license_number) != 8:
+            raise forms.ValidationError(
+                "The license number must contain 8 characters."
+            )
+
+        if not re.match(r"^[A-Z]{3}", license_number):
+            raise forms.ValidationError(
+                "The first 3 characters must be capital letters."
+            )
+
+        if not re.match(r"[0-9]{5}$", license_number):
+            raise forms.ValidationError(
+                "The last 5 characters must be numbers."
+            )
+
+        return license_number
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -6,7 +6,7 @@ from django.core.validators import RegexValidator
 from taxi.models import Driver, Car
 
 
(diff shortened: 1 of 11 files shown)


templates/taxi/car_detail.html
```diff
@@ -16,7 +16,23 @@ <h1>Drivers</h1>
...
+        <p>License number: {{ driver.license_number }}</p>
+      </li>
     {% endfor %}
   </ul>
+  
+  {% if user in car.drivers.all %}
+    <form method="post" action="{% url 'taxi:car-remove-driver' car.pk %}">
+      {% csrf_token %}
+      <button type="submit" class="btn btn-danger">Delete me from this car</button>
+    </form>
```

MNDonut (changes requested):

taxi/forms.py
```diff
@@ -0,0 +1,30 @@
+from django import forms
+from .models import Driver
```
Oleksl888: avoid using relative imports / or use get_user_model here

taxi/forms.py
```diff
@@ -0,0 +1,30 @@
+from django import forms
+from .models import Driver
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
```
Oleksl888: what about validating license on Driver Create?

taxi/forms.py
```diff
@@ -0,0 +1,30 @@
...
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    class Meta:
+        model = Driver
+        fields = ["license_number"]
+
+    def clean_license_number(self):
+        license_number = self.cleaned_data["license_number"]
+
+        if len(license_number) != 8:
```
Oleksl888: You can use RegexValidator and MaxLengthValidator here to write less code

taxi/views.py
```diff
@@ -55,16 +58,22 @@ class ManufacturerDeleteView(LoginRequiredMixin, generic.DeleteView):
 class CarListView(LoginRequiredMixin, generic.ListView):
     model = Car
     paginate_by = 5
-    queryset = Car.objects.all().select_related("manufacturer")
+    queryset = Car.objects.all().select_related(
```
Oleksl888: No need for all() here. Also why do you prefetch drivers? I don't see you using drivers on car list template

taxi/views.py
```diff
@@ -55,16 +58,22 @@ class ManufacturerDeleteView(LoginRequiredMixin, generic.DeleteView):
...
 class CarDetailView(LoginRequiredMixin, generic.DetailView):
     model = Car
+    queryset = Car.objects.prefetch_related("drivers")
 
 
 class CarCreateView(LoginRequiredMixin, generic.CreateView):
     model = Car
-    fields = "__all__"
+    fields = ["model", "manufacturer", "drivers"]
+    widgets = {
```
Oleksl888: Your code will be better structured if you add widgets on a form separately

taxi/views.py
```diff
@@ -79,11 +88,62 @@ class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
...
+        "username",
+        "first_name",
+        "last_name",
+        "license_number",
+        "password"
+    ]
+    template_name = "taxi/driver_form.html"
+    success_url = reverse_lazy("taxi:driver-list")
+
+    def form_valid(self, form):
```
Oleksl888: you will not have to do this if you inherit your Form from UserCreationForm

taxi/views.py
```diff
@@ -79,11 +88,62 @@ class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
...
+    template_name = "taxi/driver_confirm_delete.html"
+
+
+class DriverLicenseUpdateView(LoginRequiredMixin, generic.UpdateView):
+    model = Driver
+    form_class = DriverLicenseUpdateForm
+    template_name = "taxi/driver_license_update_form.html"
+    success_url = reverse_lazy("taxi:driver-detail")
+
+    def get_success_url(self):
```
Oleksl888, line 119: Here you don't need to set get_success_url because get_absolute_url is defined in your model - this will redirect to DriverDetail by itself

taxi/views.py
```diff
@@ -79,11 +88,62 @@ class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
...
+    queryset = Driver.objects.prefetch_related("cars")
 
 
 class DriverDetailView(LoginRequiredMixin, generic.DetailView):
     model = Driver
-    queryset = Driver.objects.all().prefetch_related("cars__manufacturer")
+    queryset = Driver.objects.prefetch_related("cars__manufacturer")
+
+
+@login_required
```

templates/taxi/car_detail.html
```diff
@@ -16,7 +16,23 @@ <h1>Drivers</h1>
...
+  {% else %}
+    <form method="post" action="{% url 'taxi:car-add-driver' car.pk %}">
```
Oleksl888: Keep only one form and adjust so it can set driver in CarDetailView

Oleksl888 (changes requested):

taxi/views.py
```diff
@@ -79,11 +98,47 @@ class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
...
     paginate_by = 5
+    queryset = Driver.objects.prefetch_related("cars")
 
 
 class DriverDetailView(LoginRequiredMixin, generic.DetailView):
     model = Driver
-    queryset = Driver.objects.all().prefetch_related("cars__manufacturer")
+    queryset = Driver.objects.prefetch_related("cars__manufacturer")
+
+    def post(self, request, *args, **kwargs):
```
Oleksl888: Why do you have this here?

templates/taxi/driver_license_update_form.html
Oleksl888, line 1: you don't need a special template for this - you can reuse driver_form and use yes|no tag to change label

taxi/forms.py
```diff
@@ -0,0 +1,68 @@
...
+        model = Driver
+        fields = [
+            "username",
+            "first_name",
+            "last_name",
+            "license_number",
+            "password"
+        ]
+
+    def clean_license_number(self):
```
Oleksl888: why not inherit DriverLicenseForm - then you will not have to define license validation twice

Oleksl888 (changes requested):

taxi/forms.py
```diff
@@ -0,0 +1,60 @@
...
+        model = Driver
+        fields = ["license_number"]
+
+
+class DriverLicenseUpdateForm(DriverLicenseForm):
+    class Meta(DriverLicenseForm.Meta):
+        fields = ["license_number"]
+
+
+class DriverCreateForm(DriverLicenseForm):
```
Oleksl888, line 34:
```
class DriverCreateForm(UserCreationForm):

    license_number = forms.CharField(
        validators=[
            MaxLengthValidator(8, "Maximum 8 characters allowed!"),
            RegexValidator(
                r"^[A-Z]{3}.*\d{5}$",
                "First three characters must be uppercase letters and last five characters must be digits."
            ),
        ]
    )

    class Meta:
        model = Driver
        fields = ("username", "license_number", "password", "cars")


class DriverLicenseUpdateForm(DriverCreateForm):

    class Meta:
        fields = ("license_number",)

File: templates/taxi/driver_form.html [hunk 14]
```diff
@@ -0,0 +1,53 @@
...
+          {% for field in form %}
+            {% if field.errors %}
+              <li>{{ field.label }}: {{ field.errors|striptags }}</li>
+            {% endif %}
+          {% endfor %}
+        </ul>
+      </div>
+    {% endif %}
+    
+    <form method="post">
```
Oleksl888, line 20: Why not make one template for both create/update license views and just use:
```
<h1>{{ object|yesno:"Update,Create" }} driver</h1>
  <form action="" method="post" novalidate>
    {% csrf_token %}
    {{ form|crispy }}

    <input type="submit" value="Submit" >
  </form>

Oleksl888 (approved):
Approved, but youstill can imporve - please check comments

PR #715, opened 2024-09-15
Solution

dpiuro:


dpiuro:
image
image
image
image
image
image

taxi/forms.py
```diff
@@ -0,0 +1,60 @@
+from symtable import Class
```
MNDonut: :eyes: :eyes: :eyes:

taxi/forms.py
```diff
@@ -0,0 +1,60 @@
...
+                license_number[:3].isalpha() and license_number[:3].isupper()
+        ):
+            raise forms.ValidationError(
+                "The first 3 characters must be uppercase letters."
+            )
+        if not license_number[3:].isdigit():
+            raise forms.ValidationError(
+                "The last 5 characters must be digits."
+            )
+        return license_number
```
MNDonut: create a mixin because you use the same piece of code twice

taxi/views.py
```diff
@@ -87,3 +95,45 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    model = get_user_model()
+    form_class = DriverCreationForm
+    template_name = "taxi/driver_form.html"
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+class DriverDeleteView(DeleteView):
+    model = get_user_model()
+    template_name = "taxi/driver_confirm_delete.html"
+    success_url = reverse_lazy("taxi:driver-list")\
```
MNDonut: redundant \ at the end

taxi/views.py
```diff
@@ -87,3 +95,45 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    model = get_user_model()
+    form_class = DriverLicenseUpdateForm
+    template_name = "taxi/driver_license_update_form.html"
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+@login_required
+def assign_remove_driver(request, pk):
+    car = get_object_or_404(Car, pk=pk)
+    if request.user in car.drivers.all():
```
MNDonut: use .exists() so that db uses WHERE and EXISTS at the same time because otherwise you firstly fetch 1 000 000 records and then python loops through all records to check "in" condition

templates/taxi/car_detail.html
```diff
@@ -1,22 +1,30 @@
...
+    {% empty %}
+      <li>No drivers assigned</li>
     {% endfor %}
   </ul>
+
+  <!-- Check if the current user is already a driver for this car -->
+  {% if request.user in car.drivers.all %}
+    <form method="post" action="{% url 'taxi:assign-remove-driver' car.pk %}">
+      {% csrf_token %}
+      <button type="submit" class="btn btn-danger">Remove me from this car</button>
```

MNDonut (changes requested):

taxi/forms.py
```diff
@@ -0,0 +1,41 @@
+from django import forms
+from .models import Driver, Car
+
+
+class LicenseNumberValidatorMixin:
+    def validate_license_number(self, license_number):
+        if len(license_number) != 8:
```
Oleksl888, line 30: Would be much easier to use RegexValidator instead

taxi/forms.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def clean_license_number(self):
+        license_number = self.cleaned_data.get("license_number")
+        return self.validate_license_number(license_number)
+
+
+class DriverCreationForm(forms.ModelForm):
+    class Meta:
+        model = Driver
+        fields = ["username", "license_number"]
```
Oleksl888: Driver is also a User. Should there be a password too? Also password must have some special treatment right?

taxi/forms.py
```diff
@@ -0,0 +1,41 @@
+from django import forms
+from .models import Driver, Car
```
Oleksl888: fix relative imports here

taxi/views.py
```diff
@@ -1,9 +1,17 @@
...
+    ModelForm,
+    CheckboxSelectMultiple,
+    ModelMultipleChoiceField
+)
+from django.shortcuts import render, get_object_or_404, redirect
+from django.urls import reverse_lazy, reverse
+from django.views import generic
+from django.views.generic import CreateView, DeleteView, UpdateView
 
+from .forms import DriverCreationForm, DriverLicenseUpdateForm
```
Oleksl888: change relative imports here too

taxi/views.py
```diff
@@ -87,3 +96,44 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+def assign_remove_driver(request, pk):
+    car = get_object_or_404(Car, pk=pk)
+    if car.drivers.filter(id=request.user.id).exists():
+        car.drivers.remove(request.user)
+    else:
+        car.drivers.add(request.user)
+    return redirect(reverse("taxi:car-detail", kwargs={"pk": car.pk}))
+
+
+class CarCreationForm(ModelForm):
```
Oleksl888: Form should be in the forms.py

taxi/views.py
```diff
@@ -87,3 +96,44 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+
+class DriverLicenseUpdateView(UpdateView):
+    model = get_user_model()
+    form_class = DriverLicenseUpdateForm
+    template_name = "taxi/driver_license_update_form.html"
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+@login_required
+def assign_remove_driver(request, pk):
```
Oleksl888: You can put logic to assign/remove driver into post method of CarDetailView. This way you don't need to create extra view and url endpoint.
You will need to update action on form on car detail template to send value to the cardetailview.

templates/taxi/car_detail.html
```diff
@@ -1,22 +1,29 @@
...
   <ul>
     {% for driver in car.drivers.all %}
       <li>{{ driver.username }} ({{ driver.first_name }} {{ driver.last_name }})</li>
+    {% empty %}
+      <li>No drivers assigned</li>
     {% endfor %}
   </ul>
+  
+ {% if request.user in car.drivers.all %}
+  <form method="post" action="{% url 'taxi:assign-remove-driver' car.pk %}">
```
Oleksl888: make it into one form and adjust so it can set driver in CarDetailView

templates/taxi/driver_license_update_form.html
```diff
@@ -0,0 +1,11 @@
+{% extends "base.html" %}
```
Oleksl888: you don't need to create an extra template for this. You can just use yes|no on driver_form

taxi_service_db_data.json
Oleksl888, line 1: why delete this?

Oleksl888 (changes requested):

taxi_service_db_data.json
dpiuro: Accidentally, i already have this data in db, I'll be more careful in future

Arsen-hrynevych (approved):
Overall ok

PR #717, opened 2024-09-15
first_try

yashkunn:


yashkunn:
image
image
image
image
image
image
image
image

taxi/models.py
```diff
@@ -15,7 +20,15 @@ def __str__(self):
...
 
 class Driver(AbstractUser):
-    license_number = models.CharField(max_length=255, unique=True)
+    license_number = models.CharField(
+        max_length=255,
+        unique=True,
+        validators=[
+            MinLengthValidator(8),
+            MaxLengthValidator(8),
+            RegexValidator(r"^[A-Z]{3}\d{5}$"),
```
MNDonut: do you need min and max validators if your regex checks exactly 8 characters?

templates/taxi/car_detail.html
```diff
@@ -19,4 +19,15 @@ <h1>Drivers</h1>
       <li>{{ driver.username }} ({{ driver.first_name }} {{ driver.last_name }})</li>
     {% endfor %}
   </ul>
+  {% if request.user in car.drivers.all %}
```

MNDonut (changes requested):

taxi/models.py
```diff
@@ -15,7 +20,13 @@ def __str__(self):
 
 
 class Driver(AbstractUser):
-    license_number = models.CharField(max_length=255, unique=True)
+    license_number = models.CharField(
```
Oleksl888: validation should happen on Form as per requirements

taxi/views.py
```diff
@@ -1,9 +1,10 @@
 from django.contrib.auth.decorators import login_required
-from django.shortcuts import render
+from django.shortcuts import render, get_object_or_404, redirect
 from django.urls import reverse_lazy
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
 
+from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm
```

taxi/views.py
```diff
@@ -64,18 +65,19 @@ class CarDetailView(LoginRequiredMixin, generic.DetailView):
...
     model = Car
-    fields = "__all__"
+    form_class = CarForm
     success_url = reverse_lazy("taxi:car-list")
 
 
 class CarUpdateView(LoginRequiredMixin, generic.UpdateView):
     model = Car
-    fields = "__all__"
+    form_class = CarForm
```
Oleksl888: there is no requirement to override carupdateview

taxi/views.py
```diff
@@ -64,18 +65,19 @@ class CarDetailView(LoginRequiredMixin, generic.DetailView):
...
     success_url = reverse_lazy("taxi:car-list")
 
 
 class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
     model = Car
+    queryset = Car.objects.all().prefetch_related("drivers")
```
Oleksl888: omit using .all() - prefetch_related will pull all the records anyway

taxi/views.py
```diff
@@ -87,3 +89,54 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
 class DriverDetailView(LoginRequiredMixin, generic.DetailView):
     model = Driver
     queryset = Driver.objects.all().prefetch_related("cars__manufacturer")
+
+    def get_success_url(self) -> str:
```
Oleksl888, line 108: it is not necessary to set `get_success_url` if `get_absolute_url` is defined in your model.

taxi/views.py
```diff
@@ -87,3 +89,54 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    template_name = "taxi/driver_license_update.html"
+
+    def get_success_url(self) -> str:
+        return reverse_lazy(
+            "taxi:driver-detail",
+            kwargs={"pk": self.object.pk}
+        )
+
+
+@login_required
```

templates/taxi/car_detail.html
```diff
@@ -19,4 +19,15 @@ <h1>Drivers</h1>
...
+    <form action="{% url "taxi:remove-driver" pk=car.id %}" method="post">
```
Oleksl888: Keep only one form and adjust so it can set driver in CarDetailView

Oleksl888 (changes requested):

taxi/models.py
```diff
@@ -1,4 +1,7 @@
 from django.db import models
+from django.core.validators import (
```
Oleksl888: clean up import - it is not used here anymore

taxi/forms.py
```diff
@@ -0,0 +1,52 @@
...
+        fields = (
+            UserCreationForm.Meta.fields
+            + ("first_name", "last_name", "license_number",)
+        )
+
+    def clean_license_number(self):
+        return validate_license_number(self.cleaned_data["license_number"])
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
```
Oleksl888, line 21: inherit this from DriverCreateForm and just override fields in meta so you dont have code repetition

taxi/views.py
Oleksl888: This is not done - please move this logic to the post method of your CarDetailView

templates/taxi/car_detail.html
Oleksl888: not done

Oleksl888 (changes requested):

taxi/forms.py
Nattalli: okay, but you were asked to inherit that class:
```
class DriverLicenseUpdateForm(DriverCreationForm):
```
but your solution also could be, but less responds to SOLID principles

taxi/forms.py
```diff
@@ -0,0 +1,52 @@
...
+class CarForm(forms.ModelForm):
+    drivers = forms.ModelMultipleChoiceField(
+        queryset=get_user_model().objects.all(),
+        widget=forms.CheckboxSelectMultiple,
+        required=False
+    )
+
+    class Meta:
+        model = Car
+        fields = "__all__"
```
Nattalli: don't recommend use `fields = "__all__"`

templates/taxi/car_detail.html
```diff
@@ -19,4 +19,15 @@ <h1>Drivers</h1>
...
+
+  <form method="post">
+    {% csrf_token %}
+    {% if request.user in car.drivers.all %}
+      <input type="hidden" name="action" value="Delete">
+      <button type="submit" class="btn btn-danger">Delete me from this car</button>
+    {% else %}
+      <input type="hidden" name="action" value="Add">
+      <button type="submit" class="btn btn-success">Assign me to this car</button>
+    {% endif %}
```
Nattalli, line 31: does these buttons change anything? I believe you should add new endpoints for that action

Nattalli (changes requested):

templates/taxi/car_detail.html
yashkunn: these buttons work and provide the set in the conditions of the task

taxi/forms.py
oliasymak24: You didn't change it

oliasymak24 (changes requested):

LiudmylaKulinchenko (approved):
Good job 😇

PR #719, opened 2024-09-16
commit

panichqa:


panichqa:
image
image
image
image
image
image
image

panichqa:
image
image
image
image
image

taxi/forms.py
```diff
@@ -0,0 +1,66 @@
...
+                or not license_number[:3].isupper()
+        ):
+            raise forms.ValidationError(
+                "The first 3 characters must be capital letters."
+            )
+
+        if not license_number[3:].isdigit():
+            raise forms.ValidationError(
+                "The last 5 characters must be numbers."
+            )
```
MNDonut: let's create some function for these checks because you use it twice

taxi/views.py
```diff
@@ -1,8 +1,9 @@
 from django.contrib.auth.decorators import login_required
-from django.shortcuts import render
-from django.urls import reverse_lazy
+from django.shortcuts import render, redirect, get_object_or_404
+from django.urls import reverse_lazy, reverse
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
+from taxi.forms import DriverLicenseUpdateForm, DriverCreationForm
```
MNDonut: split different groups of imports

templates/taxi/car_detail.html
```diff
@@ -1,22 +1,26 @@
...
+  <h2>Drivers</h2>
   <ul>
     {% for driver in car.drivers.all %}
       <li>{{ driver.username }} ({{ driver.first_name }} {{ driver.last_name }})</li>
     {% endfor %}
   </ul>
+
+  {% if request.user in car.drivers.all %}
+    <form action="{% url 'taxi:car-remove-driver' pk=car.id %}" method="post">
+      {% csrf_token %}
```

MNDonut (changes requested):

Nattalli (changes requested):
good job, but what about:

```

File: taxi/forms.py [hunk 4]
```diff
@@ -0,0 +1,59 @@
+from django import forms
+from django.contrib.auth.forms import UserCreationForm
+from taxi.models import Driver, Car
+
+
+def validate_license_number(license_number):
+    if len(license_number) != 8:
```
Oleksl888: You can use RegexValidator to make validation in few lines instead of making many if condidtions

taxi/forms.py
```diff
@@ -0,0 +1,59 @@
...
+            "password2"
+        )
+
+    def clean_license_number(self):
+        license_number = self.cleaned_data["license_number"]
+        validate_license_number(license_number)
+        return license_number
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
```
Oleksl888, line 28: You can use inheritance here to avoid code repetition

taxi/views.py
```diff
@@ -79,6 +81,22 @@ class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
     success_url = reverse_lazy("taxi:car-list")
 
 
+@login_required
```
Oleksl888: You can put logic to assign/remove driver into post method of CarDetailView. This way you don't need to create extra views and url endpoints.
You will need to update action on form on car detail template to send value to the cardetailview.

templates/taxi/driver_license_update_form.html
```diff
@@ -0,0 +1,11 @@
+{% extends "base.html" %}
```
Oleksl888: This template is unnecessary - driver_form is handling both create and update

templates/taxi/car_confirm_delete.html
```diff
@@ -1,10 +1,10 @@
 {% extends "base.html" %}
 
 {% block content %}
-  <h1>Delete car?</h1>
+  <h1>Delete driver?</h1>
```
Oleksl888: why driver? this is a Car view

templates/taxi/driver_list.html
```diff
@@ -1,30 +1,36 @@
...
+    {% empty %}
+      <p>No drivers available.</p>
     {% endfor %}
+  </ul>
 
-    </table>
-    {% else %}
-      <p>There are no drivers in the service.</p>
-    {% endif %}
+  <div class="pagination">
```
Oleksl888: pagination should be in the `includes` since it repeats on every list view

templates/taxi/car_detail.html
```diff
@@ -1,22 +1,26 @@
...
+      <button type="submit" class="btn btn-danger">Remove me from this car</button>
+    </form>
+  {% else %}
+    <form action="{% url 'taxi:car-add-driver' pk=car.id %}" method="post">
```
Oleksl888: make one form and adjust so it can set driver in CarDetailView

Oleksl888 (changes requested):

templates/taxi/car_confirm_delete.html
panichqa: because copy paste from driver.html :)

templates/taxi/driver_license_update_form.html
panichqa: ok, so i extends driver_form instead base.html?

Oleksl888: No, you actually can delete this template and reassign template_name on DriverLicenseUpdateView

taxi/forms.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class LicenseNumberValidationForm(forms.ModelForm):
+    def clean_license_number(self):
+        license_number = self.cleaned_data["license_number"]
+        validate_license_number(license_number)
+        return license_number
+
+
+class DriverCreationForm(UserCreationForm, LicenseNumberValidationForm):
```
Oleksl888: you can do more simple here. Instead of creating a separate class for validation define a field here:
`license_number = forms.CharField(validators=[...`

taxi/views.py
Oleksl888: This is not done - you have to group imports from django and imports from your app together and separate them with blank line

taxi/views.py
```diff
@@ -87,3 +102,32 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+
+class DriverDeleteView(LoginRequiredMixin, generic.DeleteView):
+    model = Driver
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+class DriverLicenseUpdateView(LoginRequiredMixin, generic.UpdateView):
+    model = Driver
+    form_class = DriverLicenseUpdateForm
+    template_name = "taxi/driver_license_update_form.html"
```
Oleksl888: change this reference so you can reuse the same form as DriverCreateView

templates/taxi/car_detail.html
```diff
@@ -1,22 +1,22 @@
...
   <ul>
     {% for driver in car.drivers.all %}
       <li>{{ driver.username }} ({{ driver.first_name }} {{ driver.last_name }})</li>
     {% endfor %}
   </ul>
+
+  <form method="post">
+    {% csrf_token %}
+    <input type="hidden" name="car_id" value="{{ car.id }}">
+    <button type="submit" name="action" value="{% if request.user in car.drivers.all %}remove{% else %}add{% endif %}" class="btn {% if request.user in car.drivers.all %}btn-danger{% else %}btn-success{% endif %}">
```
Oleksl888: Refactor this so you don't have to repeat this block `if request.user in car.drivers.all`

Oleksl888 (changes requested):

MNDonut (approved):

PR #720, opened 2024-09-16
Implementation to Solution of this task

Ivan-Shakhman:
solution


taxi/views.py
```diff
@@ -1,9 +1,10 @@
 from django.contrib.auth.decorators import login_required
-from django.shortcuts import render
+from django.shortcuts import render, get_object_or_404, redirect
 from django.urls import reverse_lazy
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
 
+from .forms import DriverCreateForm, DriverLicenseUpdateForm, CarForm
```
Oleksl888: use app name here to avoid relative imports

taxi/views.py
```diff
@@ -64,18 +65,19 @@ class CarDetailView(LoginRequiredMixin, generic.DetailView):
...
 class CarUpdateView(LoginRequiredMixin, generic.UpdateView):
     model = Car
-    fields = "__all__"
+    form_class = CarForm
     success_url = reverse_lazy("taxi:car-list")
 
 
 class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
     model = Car
+    form_class = CarForm
```
Oleksl888: no need to use form on delete view

taxi/views.py
```diff
@@ -64,18 +65,19 @@ class CarDetailView(LoginRequiredMixin, generic.DetailView):
...
     model = Car
-    fields = "__all__"
+    form_class = CarForm
     success_url = reverse_lazy("taxi:car-list")
 
 
 class CarUpdateView(LoginRequiredMixin, generic.UpdateView):
     model = Car
-    fields = "__all__"
+    form_class = CarForm
```
Oleksl888: there is no requirement to change form on update view

taxi/views.py
```diff
@@ -84,6 +86,39 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
     paginate_by = 5
 
 
+class DriverCreateView(LoginRequiredMixin, generic.CreateView):
+    model = Driver
+    form_class = DriverCreateForm
```
Oleksl888, line 104: what about success_url?

taxi/views.py
```diff
@@ -84,6 +86,39 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    model = Driver
+    form_class = DriverLicenseUpdateForm
+
+
 class DriverDetailView(LoginRequiredMixin, generic.DetailView):
     model = Driver
     queryset = Driver.objects.all().prefetch_related("cars__manufacturer")
+
+
+@login_required
```
Oleksl888: Try to include this logic in CarDetailView - override post method to add/remove driver to car through related fields. You will also need to adjust your form on a template to pass car_id

taxi/forms.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    class Meta(UserCreationForm.Meta):
+        model = Driver
+        fields = ("license_number",)
+
+    def clean_license_number(self):
+        license_number = self.cleaned_data["license_number"]
+        if len(license_number) != 8:
```
Oleksl888: 1. Use MaxLengthValidator and RegexValidator here
2. Instead of using clean define a field like so:
```
license_number = forms.CharField(
        validators=[

File: taxi/forms.py [hunk 7]
```diff
@@ -0,0 +1,47 @@
...
+        ):
+            raise ValidationError("First 3 characters must be uppercase chars")
+        try:
+            int(license_number[3:])
+        except ValueError:
+            raise ValidationError("last 5 characters must be a numbers")
+        return license_number
+
+
+class DriverCreateForm(UserCreationForm, DriverLicenseUpdateForm):
```
Oleksl888: Would make sense to make this a parent class being inherited from class `forms.ModelForm`. In it you add `license_number` with validators.
Then you can inherit `DriverLicenseUpdateForm` from it and only override fields as per task requirements.

taxi/forms.py
```diff
@@ -0,0 +1,47 @@
...
+    class Meta:
+        model = Driver
+        fields = UserCreationForm.Meta.fields + (
```
Oleksl888: we don't need all the fields from usercreation form here

templates/taxi/car_detail.html
```diff
@@ -3,20 +3,35 @@
...
-      Delete
-    </a>
-
-    <a href="{% url 'taxi:car-update' pk=car.id %}" class="btn btn-secondary link-to-page">
-      Update
-    </a>
   </h1>
+  {% if user.is_authenticated %}
+      {% if user not in car.drivers.all %}
+        <form action= "{% url 'taxi:add-driver' pk=car.id %}" method="post">
```
Oleksl888: This can be one form with different inputs and different values that you will later compare in your view. Also you can include a hidden field here with car_id so you can pull it out in post method in your car detail view

Oleksl888 (changes requested):

.flake8
```diff
@@ -6,5 +6,5 @@ max-complexity = 18
 select = B,C,E,F,W,T4,B9,Q0,N8,VNE
 exclude =
     **migrations
-    venv
+    .venv
```
Oleksl888: revert these changes please

taxi/forms.py
```diff
@@ -0,0 +1,42 @@
...
+from django.forms import ModelForm
+
+from taxi.models import Driver, Car
+
+
+class DriverCreateForm(forms.ModelForm):
+    license_number = forms.CharField(
+        validators=[
+            MaxLengthValidator(8),
+            RegexValidator(regex=r"^[A-Z]{3}\d{5}$")
```
Oleksl888: Adding a message to validation error is a good practice

taxi/forms.py
```diff
@@ -0,0 +1,42 @@
...
+        fields = (
+            "username",
+            "password",
+            "email",
+            "license_number",
+        )
+
+
+class DriverLicenseUpdateForm(DriverCreateForm):
+    class Meta(UserCreationForm.Meta):
```
Oleksl888: would make sense to inherit Meta from DriverCreateForm or just make it all new - usercreation meta has nothing to do here

taxi/views.py
```diff
@@ -84,6 +86,32 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
     paginate_by = 5
 
 
+class DriverCreateView(LoginRequiredMixin, generic.CreateView):
+    model = Driver
+    form_class = DriverCreateForm
+    success_url = reverse_lazy("/")
```
Oleksl888: use namespaces for redirect urls

taxi/views.py
Oleksl888: You can put logic from this view into post method of CarDetailView. We don't need to create extra view and endpoint to handle adding/deleting driver.

1. Add an empty form on car detail template that will send value 'Add' or 'Delete'
3. You override post method in CarDetailView and check values that you received in if condition (you basically slightly modify code in your function view below and put in post method).

templates/taxi/car_detail.html
```diff
@@ -3,20 +3,33 @@
...
     {{ car.model }}
-    <a href="{% url 'taxi:car-delete' pk=car.id %}" class="btn btn-danger link-to-page">
-      Delete
-    </a>
-
-    <a href="{% url 'taxi:car-update' pk=car.id %}" class="btn btn-secondary link-to-page">
-      Update
-    </a>
   </h1>
+  {% if car in user.cars.all %}
```
Oleksl888: maybe refactor this to if `user in car.drivers.all` and add prefetch on the view to fix n+1 problem

Oleksl888 (changes requested):

templates/taxi/driver_license_form.html
```diff
@@ -0,0 +1,8 @@
+{% extends "base.html" %}
+{% block content %}
+  <form action="" method="post">
+    {% csrf_token %}
+
+    <input type="submit" value="Yes" class="btn btn-danger">
+  </form>
+{% endblock %}
```
Nattalli: add empty line

templates/taxi/driver_form.html
```diff
@@ -0,0 +1,12 @@
...
+{% block content %}
+  <h1>Registration Driver:</h1>
+  <form action="" method="post">
+    {% csrf_token %}
+    {{ form|crispy }}
+    <br>
+    <input type="submit" value="apply">
+
+  </form>
+{% endblock %}
```
Nattalli: add empty line

templates/taxi/driver_confirm_delete.html
```diff
@@ -0,0 +1,11 @@
...
+
+{% block content %}
+  <h1>Delete Driver {{ driver.username }}?</h1>
+  <form action="" method="post">
+    {% csrf_token %}
+
+    <input type="submit" value="Yes" class="btn btn-danger">
+    <a href="{% url 'taxi:driver-detail' pk=driver.id %}" class="btn btn-secondary link-to-page">No</a>
+  </form>
+{% endblock %}
```
Nattalli: add empty line

taxi/forms.py
```diff
@@ -0,0 +1,50 @@
...
+from pyexpat.errors import messages
+
+from taxi.models import Driver, Car
+
+
+class DriverCreateForm(forms.ModelForm):
+    license_number = forms.CharField(
+        validators=[
+            MaxLengthValidator(
+                8,
```
Nattalli: make 8 as a constant

Nattalli (changes requested):
did you forget about that point from the task?

```

**Comment by Nattalli** [comment]
also: **NOTE: Attach screenshots of all created or modified pages to pull request. It's important to attach images not links to them.**

**Comment by Ivan-Shakhman** [comment]
image


image

File: taxi/forms.py [hunk 19]
```diff
@@ -0,0 +1,51 @@
+from django import forms
+from django.contrib.auth import get_user_model
+from django.contrib.auth.forms import UserCreationForm
+from django.core.validators import MaxLengthValidator, RegexValidator
+from django.forms import ModelForm
+from pyexpat.errors import messages
+
+from taxi.models import Driver, Car
+LIMITED_LENGTH = 8
```
Arsen-hrynevych, line 9: add 2 blank lines

taxi/views.py
```diff
@@ -1,10 +1,13 @@
...
-from django.shortcuts import render
+from django.http import HttpResponseRedirect
+from django.shortcuts import render, get_object_or_404, redirect
 from django.urls import reverse_lazy
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
 
-from .models import Driver, Car, Manufacturer
+from taxi.forms import DriverCreateForm, DriverLicenseUpdateForm, CarForm
+from taxi.models import Driver, Car, Manufacturer
```
Arsen-hrynevych, line 10: If importing more than two items, it is better to write them on separate lines. This improves readability and maintenance of the code. 

Example:

From taxi.models import ( 
	Driver,
	Manufacturer,
	Car
)
Read more about topic [here](https://peps.python.org/pep-0008/#imports)

Arsen-hrynevych (approved):
Thanks

PR #721, opened 2024-09-16
Task Solution

achalaidiuk:


taxi/forms.py
```diff
@@ -0,0 +1,47 @@
+from django.contrib.auth.forms import UserCreationForm
+from django import forms
+
+from taxi.models import Driver, Car
+
+
+class CleanLicenseNumberMixin:
+    def clean_license_number(self):
```
Oleksl888: you can simplify this by using MaxLengthValidator and RegexValidator on license_number field in a form

taxi/forms.py
```diff
@@ -0,0 +1,47 @@
...
+
+class DriverCreationForm(CleanLicenseNumberMixin, UserCreationForm):
+    class Meta:
+        model = Driver
+        fields = UserCreationForm.Meta.fields + (
+            "license_number", "first_name", "last_name"
+        )
+
+
+class DriverLicenseUpdateForm(CleanLicenseNumberMixin, forms.ModelForm):
```
Oleksl888: you can inherit this Form from above - and so you will have to only override fields in Meta class

taxi/views.py
```diff
@@ -1,9 +1,13 @@
...
+
 from django.contrib.auth.decorators import login_required
-from django.shortcuts import render
+from django.http import HttpRequest, HttpResponse
+from django.shortcuts import render, redirect
 from django.urls import reverse_lazy
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
 
+from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm
```

taxi/views.py
```diff
@@ -1,9 +1,13 @@
+from http.client import HTTPResponse
```
Oleksl888: we use HttpResponse from django here

taxi/views.py
```diff
@@ -87,3 +108,15 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
 class DriverDetailView(LoginRequiredMixin, generic.DetailView):
     model = Driver
     queryset = Driver.objects.all().prefetch_related("cars__manufacturer")
+
+
+@login_required
+def assign_car(request: HttpRequest, pk: int) -> HttpResponse:
```
Oleksl888: You can put this logic into `post` method of `CarDetailView`. The reason is that we don't need to create extra view and url endpoint to handle adding/deleting driver.

2. You override post method in this view and catch those values that you received in if condition and add/remove driver to car through related fields (basically you can copy logic from this method over there).

templates/taxi/car_detail.html
```diff
@@ -19,4 +18,19 @@ <h1>Drivers</h1>
       <li>{{ driver.username }} ({{ driver.first_name }} {{ driver.last_name }})</li>
     {% endfor %}
   </ul>
+  {% if user.is_authenticated and user in car.drivers.all %}
```

templates/taxi/driver_list.html
```diff
@@ -1,29 +1,46 @@
 {% extends "base.html" %}
 
 {% block content %}
-    <h1>Driver List
+    <h1>
```
Oleksl888: check indentation everywhere - needs to be 2 spaces in html

Oleksl888 (changes requested):

taxi/models.py
```diff
@@ -15,7 +16,17 @@ def __str__(self):
 
 
 class Driver(AbstractUser):
-    license_number = models.CharField(max_length=255, unique=True)
+    license_number = models.CharField(
+        blank=True,
+        max_length=8,
+        validators=[
+            MaxLengthValidator(8),
```
Oleksl888: validation should happen on  a form in this exercise

templates/base.html
```diff
@@ -6,35 +6,29 @@
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <link rel="stylesheet"
-      href="https://cdn.jsdelivr.net/npm/bootstrap@4.5.3/dist/css/bootstrap.min.css"
-      integrity="sha384-TX8t27EcRE3e/ihU7zmQxVncDAy5uIKz4rEkgIXeMed4M0jlfIDPvg6uqKI2xXr2"
-      crossorigin="anonymous">
-  <!-- Add additional CSS in static file -->
+        href="https://cdn.jsdelivr.net/npm/bootstrap@4.5.3/dist/css/bootstrap.min.css"
```
Oleksl888: check indentation

taxi/views.py
```diff
@@ -87,3 +124,15 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    car = Car.objects.get(pk=pk)
+
+    if request.user in car.drivers.all():
```
Oleksl888: clean this up - you are not using this view anymore

Oleksl888 (changes requested):

taxi/views.py
```diff
@@ -61,10 +63,28 @@ class CarListView(LoginRequiredMixin, generic.ListView):
 class CarDetailView(LoginRequiredMixin, generic.DetailView):
     model = Car
 
+    def get_queryset(self):
+        return Car.objects.prefetch_related("drivers").all()
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["is_driver"] = self.request.user in self.object.drivers.all()
```
MNDonut: So. In order to do "in" operator you call "self.object.drivers.all()". The order of operations is:
1. self.object.drivers.all() - django fetches drivers. They aren't prefetched, thus 2 SQL queries are called.
2. Django fetched all records from db
3. Python runs "in" operator that will go through all records if there is a single object

You can optimize this on ORM lvl so you don't need to use "in" operator. Let's do it
The same for 77 line in "post" method

MNDonut (changes requested):

taxi/views.py
```diff
@@ -61,10 +63,30 @@ class CarListView(LoginRequiredMixin, generic.ListView):
 class CarDetailView(LoginRequiredMixin, generic.DetailView):
     model = Car
 
+    def get_queryset(self):
+        return Car.objects.prefetch_related("drivers").all()
```
Oleksl888: No need for .all() in the end - prefetch pulls all records already

taxi/forms.py
```diff
@@ -0,0 +1,71 @@
...
+                        " of 3 uppercase letters followed by 5 digits."
+            ),
+        ]
+    )
+    first_name = forms.CharField(max_length=30)
+    last_name = forms.CharField(max_length=30)
+
+    def clean_license_number(self):
+        license_number = self.cleaned_data["license_number"]
+        if len(license_number) != 8:
```
Oleksl888: you already have `max_length=8` on a form so this method is unnecessary. What you could do is to add MaxValueValidator to the validators attribute to make your code better structured

taxi/forms.py
```diff
@@ -0,0 +1,71 @@
...
+        return license_number
+
+    class Meta:
+        model = Driver
+        fields = UserCreationForm.Meta.fields + (
+            "license_number", "first_name", "last_name"
+        )
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
```
Oleksl888: Try to inherit this from DriverCreationForm and only redefine Meta Class

Oleksl888 (changes requested):

MNDonut (approved):

PR #722, opened 2024-09-16
Solution

pashawarganov:
Car detail:
image
Driver list:
image
Driver detail:
image
Driver delete:
image
Driver update:
image
Driver create:
image


taxi/forms.py
```diff
@@ -0,0 +1,47 @@
...
+from taxi.models import Driver, Car
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    class Meta:
+        model = Driver
+        fields = ("license_number",)
+
+    def clean_license_number(self):
+        license_number = self.cleaned_data["license_number"]
```
Oleksl888: you can simplify this by using MaxLengthValidator and RegexValidator on license_number field

taxi/views.py
```diff
@@ -64,13 +66,13 @@ class CarDetailView(LoginRequiredMixin, generic.DetailView):
...
     model = Car
-    fields = "__all__"
+    form_class = CarForm
     success_url = reverse_lazy("taxi:car-list")
 
 
 class CarUpdateView(LoginRequiredMixin, generic.UpdateView):
     model = Car
-    fields = "__all__"
+    form_class = CarForm
```
Oleksl888: we don't need to change form on update view

taxi/views.py
```diff
@@ -87,3 +89,37 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    success_url = reverse_lazy("taxi:driver-list")
+    form_class = DriverLicenseUpdateForm
+
+
+class DriverDeleteView(LoginRequiredMixin, generic.DeleteView):
+    model = Driver
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+def remove_driver_from_car(request, pk):
```
Oleksl888: You can put logic from `remove_driver_from_car` and `assign_driver_to_car` into post method of `CarDetailView`. The reason is that we don't need to create two extra views and two extra endpoints to handle adding/deleting driver.

1. Add an empty form instead of links on car detail template that will send value 'Add' or 'Delete'
3. You override post method in this view and catch those values that you received in if condition and add/remove driver to car through related fields.
4. You will need to figure out how to get car and user in the view's post method to set related fields.

templates/taxi/car_detail.html
```diff
@@ -1,22 +1,38 @@
...
-    <a href="{% url 'taxi:car-update' pk=car.id %}" class="btn btn-secondary link-to-page">
-      Update
-    </a>
   </h1>
   <p>Manufacturer: ({{ car.manufacturer.name }}, {{ car.manufacturer.country }})</p>
-  <h1>Drivers</h1>
-  <hr>
+  <h3>
+    Drivers
+    {% if request.user in car.drivers.all %}
```
Oleksl888: need to fix n+1 problem here and change the below block of code so it sends post request to cardetailview instead

Oleksl888 (changes requested):

taxi/views.py
```diff
@@ -55,22 +57,38 @@ class ManufacturerDeleteView(LoginRequiredMixin, generic.DeleteView):
...
 class CarCreateView(LoginRequiredMixin, generic.CreateView):
     model = Car
-    fields = "__all__"
+    form_class = CarForm
     success_url = reverse_lazy("taxi:car-list")
 
 
 class CarUpdateView(LoginRequiredMixin, generic.UpdateView):
     model = Car
-    fields = "__all__"
```
Oleksl888, line 73: you should keep the fields here

taxi/views.py
```diff
@@ -55,22 +57,38 @@ class ManufacturerDeleteView(LoginRequiredMixin, generic.DeleteView):
 class CarListView(LoginRequiredMixin, generic.ListView):
     model = Car
     paginate_by = 5
-    queryset = Car.objects.all().select_related("manufacturer")
+    queryset = (
+        Car.objects.all()
+        .select_related("manufacturer")
+        .prefetch_related("drivers")
```
Oleksl888: you need prefetch_related on CarDetailView, not here.

Oleksl888 (changes requested):

taxi/views.py
```diff
@@ -60,12 +60,15 @@ class CarListView(LoginRequiredMixin, generic.ListView):
...
-        .prefetch_related("drivers")
     )
 
 
 class CarDetailView(LoginRequiredMixin, generic.DetailView):
     model = Car
+    queryset = (
+        Car.objects.all()
+        .prefetch_related("drivers")
+    )
```
MNDonut: .all() is redundant, every method calls it automatically

MNDonut (changes requested):

Oleksl888 (approved):
Well done! :saluting_face:

PR #723, opened 2024-09-16
Solution

mihavryliuk:

### DIFF
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -1,0 +1,65 @@
+from django.contrib.auth import get_user_model
+from django.contrib.auth.forms import UserCreationForm
+from django import forms
+
+from taxi.models import Driver, Car
+
+
+class DriverCreationForm(UserCreationForm):
+    class Meta(UserCreationForm.Meta):
+        model = Driver
+        fields = UserCreationForm.Meta.fields + ("license_number",)
+
+    def clean_license_number(self):
+        license_number = self.cleaned_data["license_number"]
+        if len(license_number) != 8:
+            raise forms.ValidationError(
+                "License number must be exactly 8 characters long."
+            )
+        if (not license_number[:3].isalpha()
+                or not license_number[:3].isupper()):
+            raise forms.ValidationError(
+                "The first three characters must be uppercase letters."
+            )
+        if not license_number[3:].isdigit():
+            raise forms.ValidationError(
+                "The last five characters must be digits."
+            )
+
+        return license_number
+
+
+class CarForm(forms.ModelForm):
+    drivers = forms.ModelChoiceField(
+        queryset=get_user_model().objects.all(),
+        widget=forms.CheckboxSelectMultiple(),
+        required=False,
+    )
+
+    class Meta:
+        model = Car
+        fields = "__all__"
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    class Meta:
+        model = Driver
+        fields = ("license_number",)
+
+    def clean_license_number(self):
+        license_number = self.cleaned_data["license_number"]
+        if len(license_number) != 8:
+            raise forms.ValidationError(
+                "License number must be exactly 8 characters long."
+            )
+        if (not license_number[:3].isalpha()
+                or not license_number[:3].isupper()):
+            raise forms.ValidationError(
+                "The first three characters must be uppercase letters."
+            )
+        if not license_number[3:].isdigit():
+            raise forms.ValidationError(
+                "The last five characters must be digits."
+            )
+
+        return license_number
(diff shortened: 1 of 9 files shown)


taxi/forms.py
```diff
@@ -0,0 +1,65 @@
...
+from taxi.models import Driver, Car
+
+
+class DriverCreationForm(UserCreationForm):
+    class Meta(UserCreationForm.Meta):
+        model = Driver
+        fields = UserCreationForm.Meta.fields + ("license_number",)
+
+    def clean_license_number(self):
+        license_number = self.cleaned_data["license_number"]
```
Oleksl888: you can simplify this by using built-in Validators on license_number field

taxi/forms.py
```diff
@@ -0,0 +1,65 @@
...
+        required=False,
+    )
+
+    class Meta:
+        model = Car
+        fields = "__all__"
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    class Meta:
```
Oleksl888, line 46: apply inheritance here so you don't need to copy all this code

taxi/views.py
```diff
@@ -1,9 +1,11 @@
 from django.contrib.auth.decorators import login_required
-from django.shortcuts import render
+from django.shortcuts import render, redirect
 from django.urls import reverse_lazy
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
+from django.views.generic import UpdateView
 
+from .forms import DriverCreationForm, CarForm, DriverLicenseUpdateForm
```
Oleksl888: avoid using relative import

taxi/views.py
```diff
@@ -61,10 +63,18 @@ class CarListView(LoginRequiredMixin, generic.ListView):
 class CarDetailView(LoginRequiredMixin, generic.DetailView):
     model = Car
 
+    def post(self, request, *args, **kwargs):
```
Oleksl888, line 71: well done! :clap:

templates/taxi/car_detail.html
```diff
@@ -19,4 +19,19 @@ <h1>Drivers</h1>
       <li>{{ driver.username }} ({{ driver.first_name }} {{ driver.last_name }})</li>
     {% endfor %}
   </ul>
+  
+  <hr>
+  {% if user in car.drivers.all %}
```
Oleksl888, line 24: fix N+1 problem here

templates/taxi/car_detail.html
```diff
@@ -19,4 +19,19 @@ <h1>Drivers</h1>
...
+    <form method="post" action="{% url 'taxi:car-detail' pk=car.id %}">
```
Oleksl888, line 25: this can be one form - use condition to show assign/remove inputs
Also you can leave action empty since you post to the same view

Oleksl888 (changes requested):

taxi/urls.py
```diff
@@ -9,10 +9,12 @@
     CarDeleteView,
     DriverListView,
     DriverDetailView,
+    DriverCreateView,
+    DriverDeleteView,
     ManufacturerListView,
     ManufacturerCreateView,
     ManufacturerUpdateView,
-    ManufacturerDeleteView,
+    ManufacturerDeleteView, DriverLicenseUpdateView,
```
Oleksl888, line 17: this should start at new line

taxi/forms.py
Oleksl888: Something like this so you dont have to enter license_number twice:
```
class DriverLicenseUpdateForm(DriverCreationForm):

    class Meta:
        fields = ("license_number",)

File: taxi/views.py [hunk 8]
```diff
@@ -55,16 +57,29 @@ class ManufacturerDeleteView(LoginRequiredMixin, generic.DeleteView):
 class CarListView(LoginRequiredMixin, generic.ListView):
     model = Car
     paginate_by = 5
-    queryset = Car.objects.all().select_related("manufacturer")
+    queryset = (
+        Car.objects
+        .all()
```
Oleksl888, line 62: you dont need all() here since you have select_related

taxi/views.py
```diff
@@ -55,16 +57,29 @@ class ManufacturerDeleteView(LoginRequiredMixin, generic.DeleteView):
...
+        .select_related("manufacturer")
+        .prefetch_related("drivers")
```
Oleksl888, line 64: you need to prefetch drivers on CarDetail - because that's where  you access that field

templates/taxi/car_detail.html
Oleksl888: One form is enough here - something like this:
```
<form method="post">
    {% csrf_token %}
    {{ form }}
    {% if user in car.drivers.all %}
    <input type="submit" value="remove">
    {% else %}
    <input type="submit" value="assign">
    {% endif %}
  </form>

Oleksl888 (approved):
Approved - but please check the comments and fix

PR #724, opened 2024-09-16
Develop

vladislav-tsybuliak1:
user_form_driver_1 (create on the list)
user_form_driver_2 (create form with validation)
user_form_driver_3 (driver detail)
user_form_driver_4 (driver update with validation)
user_form_driver_5 (driver deletion)
user_form_driver_6 (car creation)
user_form_driver_7 (assign me)
user_form_driver_8 (delete me)


taxi/forms.py
```diff
@@ -0,0 +1,49 @@
...
+    class Meta(UserCreationForm.Meta):
+        model = get_user_model()
+        fields = UserCreationForm.Meta.fields + (
+            "first_name",
+            "last_name",
+            "license_number",
+        )
+
+
+class DriverLicenseUpdateForm(LicenseNumberMixin, forms.ModelForm):
```
Oleksl888: using inheritance here would be a better approach (add mixin is also possible, but in this case UpdateForm is exactly the same as CreateForm and only fields in Meta are different)

taxi/views.py
```diff
@@ -1,17 +1,20 @@
...
-from django.shortcuts import render
-from django.urls import reverse_lazy
+from django.http import HttpRequest, HttpResponse
+from django.shortcuts import render, get_object_or_404, redirect
+from django.urls import reverse_lazy, reverse
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
 
-from .models import Driver, Car, Manufacturer
+from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm
```

taxi/views.py
```diff
@@ -64,14 +67,18 @@ class CarDetailView(LoginRequiredMixin, generic.DetailView):
...
+
+    def get_success_url(self) -> str:
+        return reverse("taxi:car-detail", kwargs={"pk": self.object.id})
 
 
 class CarUpdateView(LoginRequiredMixin, generic.UpdateView):
     model = Car
-    fields = "__all__"
-    success_url = reverse_lazy("taxi:car-list")
+    form_class = CarForm
```
Oleksl888: there is no requirement to use this form on update view

taxi/views.py
```diff
@@ -80,10 +87,41 @@ class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
...
+    model = get_user_model()
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+class DriverUpdateLicenseView(LoginRequiredMixin, generic.UpdateView):
+    model = get_user_model()
+    form_class = DriverLicenseUpdateForm
+
+
+@login_required
```
Oleksl888: You can put this logic into post method of CarDetailView. The reason is that we don't need to create extra views and url endpoints to handle adding/deleting driver.

templates/taxi/car_detail.html
```diff
@@ -3,20 +3,37 @@
...
-    <a href="{% url 'taxi:car-update' pk=car.id %}" class="btn btn-secondary link-to-page">
-      Update
-    </a>
   </h1>
   <p>Manufacturer: ({{ car.manufacturer.name }}, {{ car.manufacturer.country }})</p>
-  <h1>Drivers</h1>
+  <h1>
+    Drivers
+  </h1>
+  {% if user in car.drivers.all %}
```
Adjust the form so it can set driver in CarDetailView (you will need only one form, not two)

Oleksl888 (changes requested):

Oleksl888 (approved):
Good job!

PR #726, opened 2024-09-17
Resolve

ihorutkin:


ihorutkin:
image
image
image
image
image

taxi/forms.py
```diff
@@ -0,0 +1,48 @@
+from django.contrib.auth import get_user_model
+from django.contrib.auth.forms import UserCreationForm
+from django.core.exceptions import ValidationError
+
+from taxi.models import Driver, Car
+from django import forms
+
+
+class DriverCreationForm(UserCreationForm):
+    class Meta(UserCreationForm.Meta):
```
Oleksl888, line 30: this should have license number validation as well. Also think how to apply inheritance so you define your forms more efficiently

taxi/forms.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    length = 8
+
+    class Meta:
+        model = Driver
+        fields = ("license_number",)
+
+    def clean_license_number(self):
```
Oleksl888: you can simplify this by using MaxLengthValidator and RegexValidator on license_number field in a form

taxi/views.py
```diff
@@ -61,16 +65,47 @@ class CarListView(LoginRequiredMixin, generic.ListView):
...
+        context = super().get_context_data(**kwargs)
+        car = self.object
+        user = self.request.user
+
+        if isinstance(user, Driver):
+            context["is_driver"] = car.drivers.filter(pk=user.id).exists()
+        return context
+
+
+class CarAssignView(LoginRequiredMixin, generic.UpdateView):
```

taxi/views.py
```diff
@@ -61,16 +65,47 @@ class CarListView(LoginRequiredMixin, generic.ListView):
...
     model = Car
-    fields = "__all__"
+    form_class = CarForm
     success_url = reverse_lazy("taxi:car-list")
 
 
 class CarUpdateView(LoginRequiredMixin, generic.UpdateView):
     model = Car
-    fields = "__all__"
+    form_class = CarForm
```
Oleksl888: no need to change form here as per requirements

templates/taxi/car_detail.html
```diff
@@ -13,6 +13,14 @@ <h1>
   </h1>
   <p>Manufacturer: ({{ car.manufacturer.name }}, {{ car.manufacturer.country }})</p>
   <h1>Drivers</h1>
+  <form method="post" action="{% if is_driver %}{% url 'taxi:driver-resign' car.pk %}{% else %}{% url 'taxi:driver-assign' car.pk %}{% endif %}">
```
Oleksl888: Adjust the form so it can set driver in CarDetailView

taxi/views.py
```diff
@@ -1,9 +1,13 @@
...
+
 from django.contrib.auth.decorators import login_required
-from django.shortcuts import render
+from django.http import HttpResponseForbidden
+from django.shortcuts import render, redirect, get_object_or_404
 from django.urls import reverse_lazy
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
 
+from .forms import DriverCreationForm, DriverLicenseUpdateForm, CarForm
```

Oleksl888 (changes requested):

taxi/views.py
```diff
@@ -61,10 +61,32 @@ class CarListView(LoginRequiredMixin, generic.ListView):
...
+        if isinstance(user, Driver):
+            context["is_driver"] = car.drivers.filter(pk=user.id).exists()
+        return context
+
+    def post(self, request, *args, **kwargs):
+        car = self.get_object()
+        driver = self.request.user
+        action = request.POST.get("action")
+
+        if isinstance(driver, Driver):
```
Oleksl888, line 76: you don't need to check this - with current implementation driver = user and this view is availlable only for logged in users

templates/taxi/car_detail.html
```diff
@@ -13,6 +13,19 @@ <h1>
   </h1>
   <p>Manufacturer: ({{ car.manufacturer.name }}, {{ car.manufacturer.country }})</p>
   <h1>Drivers</h1>
+  {% if is_driver %}
+    <p>You are already assigned to this car.</p>
+    <form action="" method="post">
```
Oleksl888, line 16: you can use only one form with condition to show one button or another

templates/taxi/driver_license_updating.html
Oleksl888, line 1: this template is unnecessary - it just duplicates driver_form above

Oleksl888 (changes requested):

LiudmylaKulinchenko (approved):
Good job 🌸

PR #728, opened 2024-09-17
task done

Sergi0bbb:
button that leads to the driver creation page:
image

driver update page:
image

delete and update buttons:
image

checkboxes on car creation page:
image

buttons Assign me to this car and Delete me from this car
image
image

### DIFF
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -2,6 +2,7 @@ import re
 
 from django import forms
 from django.contrib.auth import get_user_model
+from django.contrib.auth.forms import UserCreationForm
 from django.core.exceptions import ValidationError
 
 from taxi.models import Driver, Car
@@ -26,8 +27,6 @@ class DriverLicenseUpdateForm(forms.ModelForm):
     class Meta:
         model = Driver
         fields = (
-            "last_name",
-            "first_name",
             "license_number",
         )
 
@@ -35,20 +34,13 @@ class DriverLicenseUpdateForm(forms.ModelForm):
         return validate_license_number(self.cleaned_data.get("license_number"))
 
 
-class DriverLicenseCreateForm(forms.ModelForm):
-    cars = forms.ModelMultipleChoiceField(
-        queryset=Car.objects.all(),
-        widget=forms.CheckboxSelectMultiple
-    )
-
+class DriverCreateForm(UserCreationForm):
     class Meta:
         model = Driver
-        fields = (
-            "username",
+        fields = UserCreationForm.Meta.fields + (
             "first_name",
             "last_name",
             "license_number",
-            "cars"
         )
 
     def clean_license_number(self):
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -23,18 +23,18 @@ def validate_license_number(license_number):
     return license_number
 
 
-class DriverLicenseUpdateForm(forms.ModelForm):
+class LicenseFormMixin(forms.ModelForm):
+    license_number = forms.CharField(
+        max_length=8,
+        required=True,
+        validators=[validate_license_number]
+    )
+
     class Meta:
-        model = Driver
-        fields = (
-            "license_number",
-        )
-
-    def clean_license_number(self):
-        return validate_license_number(self.cleaned_data.get("license_number"))
+        abstract = True
 
 
-class DriverCreateForm(UserCreationForm):
+class DriverCreateForm(LicenseFormMixin, UserCreationForm):
     class Meta:
         model = Driver
         fields = UserCreationForm.Meta.fields + (
@@ -43,8 +43,11 @@ class DriverCreateForm(UserCreationForm):
             "license_number",
         )
 
-    def clean_license_number(self):
-        return validate_license_number(self.cleaned_data.get("license_number"))
+
+class DriverLicenseUpdateForm(LicenseFormMixin, forms.ModelForm):
+    class Meta:
+        model = Driver
+        fields = ("license_number",)
 
 
 class CarForm(forms.ModelForm):
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -1,0 +1,66 @@
+import re
+
+from django import forms
+from django.contrib.auth import get_user_model
+from django.core.exceptions import ValidationError
+
+from taxi.models import Driver, Car
+
+
+def validate_license_number(license_number):
+    if len(license_number) != 8:
+        raise ValidationError("License number length must be 8")
+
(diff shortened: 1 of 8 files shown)


taxi/views.py
```diff
@@ -1,10 +1,12 @@
 from django.contrib.auth.decorators import login_required
-from django.shortcuts import render
+from django.shortcuts import render, get_object_or_404, redirect
 from django.urls import reverse_lazy
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
 
-from .models import Driver, Car, Manufacturer
+from taxi.forms import DriverLicenseUpdateForm, DriverLicenseCreateForm, \
```
MNDonut: we don't use \ as a line breaker

templates/taxi/car_detail.html
```diff
@@ -19,4 +19,14 @@ <h1>Drivers</h1>
       <li>{{ driver.username }} ({{ driver.first_name }} {{ driver.last_name }})</li>
     {% endfor %}
   </ul>
+  {% if user.is_authenticated %}
+    <form action="{% url 'taxi:assign-or-delete-driver-to-car' pk=car.pk %}" method="post">
+      {% csrf_token %}
+      {% if user in car.drivers.all %}
+        <button type="submit" class="btn btn-danger">Delete me from this car</button>
+      {% else %}
```

MNDonut (changes requested):

taxi/forms.py
```diff
@@ -0,0 +1,66 @@
...
+        raise ValidationError("Last 5 characters must be digits")
+
+    return license_number
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    class Meta:
+        model = Driver
+        fields = (
+            "last_name",
```
Oleksl888: Update form should update only drivers license

taxi/forms.py
```diff
@@ -0,0 +1,66 @@
...
+            "first_name",
+            "license_number",
+        )
+
+    def clean_license_number(self):
+        return validate_license_number(self.cleaned_data.get("license_number"))
+
+
+class DriverLicenseCreateForm(forms.ModelForm):
```
Oleksl888: why is it called DriverLicenseCreateForm?

taxi/forms.py
```diff
@@ -0,0 +1,66 @@
...
+    cars = forms.ModelMultipleChoiceField(
```
Oleksl888: Whats the point to include cars here? These relations will not be saved in generic CreateView because it's a backwards relationship - you will need to add logic in view to save m2m relations then

taxi/forms.py
```diff
@@ -0,0 +1,66 @@
...
+        queryset=Car.objects.all(),
+        widget=forms.CheckboxSelectMultiple
+    )
+
+    class Meta:
+        model = Driver
```
Oleksl888, line 39: Driver is also a User Model - how will you handle password saving?

taxi/views.py
```diff
@@ -61,16 +66,19 @@ class CarListView(LoginRequiredMixin, generic.ListView):
...
     model = Car
-    fields = "__all__"
+    form_class = CarForm
     success_url = reverse_lazy("taxi:car-list")
 
 
 class CarUpdateView(LoginRequiredMixin, generic.UpdateView):
     model = Car
-    fields = "__all__"
+    form_class = CarForm
```
Oleksl888: there is no requirement to change form on update view

taxi/views.py
```diff
@@ -87,3 +95,30 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
 class DriverDetailView(LoginRequiredMixin, generic.DetailView):
     model = Driver
     queryset = Driver.objects.all().prefetch_related("cars__manufacturer")
+
+
+class AssignOrDeleteUserToCarView(LoginRequiredMixin, generic.View):
```
Oleksl888: You can put this logic into post method of CarDetailView. This way you don't need to create extra view and url endpoint to handle adding/deleting driver.
You will need to update action on  form on car detail template to send value to the cardetailview.

templates/taxi/car_detail.html
```diff
@@ -19,4 +19,14 @@ <h1>Drivers</h1>
       <li>{{ driver.username }} ({{ driver.first_name }} {{ driver.last_name }})</li>
     {% endfor %}
   </ul>
+  {% if user.is_authenticated %}
+    <form action="{% url 'taxi:assign-or-delete-driver-to-car' pk=car.pk %}" method="post">
```
Oleksl888: adjust form so it can set driver in CarDetailView

Oleksl888 (changes requested):

taxi/views.py
```diff
@@ -87,3 +117,30 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    def post(self, request, pk):
+        car = get_object_or_404(Car, pk=pk)
```
Oleksl888: remove this view - since you redefined CarDetailView

taxi/forms.py
```diff
@@ -0,0 +1,58 @@
...
+            "First 3 characters must be uppercase letters"
+        )
+
+    if not license_number[3:].isdigit():
+        raise ValidationError("Last 5 characters must be digits")
+
+    return license_number
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
```
Oleksl888: you can inherit this from DriverCreateForm to avoid code duplication

taxi/urls.py
```diff
@@ -46,6 +50,22 @@
...
+        "drivers/<int:pk>/delete/",
+        DriverDeleteView.as_view(),
+        name="driver-delete"
+    ),
+    path(
+        "drivers/<int:pk>/update/",
+        DriverUpdateView.as_view(),
+        name="driver-update"
+    ),
+    path(
```
Oleksl888: clean up this since you changed the logic for assigning driver

taxi/views.py
```diff
@@ -61,10 +67,34 @@ class CarListView(LoginRequiredMixin, generic.ListView):
 class CarDetailView(LoginRequiredMixin, generic.DetailView):
     model = Car
 
+    def post(self, request, *args, **kwargs):
+        action = request.POST.get("action")
+        pk = kwargs["pk"]
```
Oleksl888: as an alternative you can use `self.get_object()` - on DetailView it will pull the instance that we are working with so you will need to pass pk ina form

Oleksl888 (changes requested):

Oleksl888 (approved):
Approved!

PR #730, opened 2024-09-17
first

egor250330:


taxi/forms.py
```diff
@@ -0,0 +1,48 @@
+from django.contrib.auth import get_user_model
+from django.contrib.auth.forms import UserCreationForm
+from django.core.exceptions import ValidationError
+
+from taxi.models import Driver, Car
+from django import forms
```
MNDonut: fix imports ordering

taxi/forms.py
```diff
@@ -0,0 +1,48 @@
...
+        license_num = self.cleaned_data["license_number"]
+        first_char = license_num[:3]
+        last_char = license_num[3:]
+
+        if (len(license_num) != DriverLicenseUpdateForm.length
+                or first_char.upper() != first_char
+                or not first_char.isalpha()
+                or not last_char.isnumeric()):
+            raise ValidationError("Data is not correct")
+        return license_num
```
MNDonut: too hard to read, too many != and "not" comparisons
it's better either use "not" before the whole statement of do like this:
"if a and b and c": return num
else: raise error

taxi/urls.py
```diff
@@ -42,10 +48,31 @@
...
+        CarAssignView.as_view(),
+        name="driver-assign"
+    ),
+    path(
+        "cars/<int:pk>/resign/",
+        CarResignView.as_view(),
+        name="driver-resign"
+    ),
     path("drivers/", DriverListView.as_view(), name="driver-list"),
+    path("drivers/create", DriverCreateView.as_view(), name="driver-create"),
```
MNDonut: / at the end

taxi/urls.py
```diff
@@ -42,10 +48,31 @@
...
     ),
+    path(
+        "drivers/license/<int:pk>/",
+        DriverUpdateView.as_view(),
+        name="driver-update"
+    ),
+    path(
+        "drivers/delete/<int:pk>/",
+        DriverDeleteView.as_view(),
+        name="driver-delete"
```
MNDonut: drivers/id/... - it's a recommended format so that it's easier to manage urls

taxi/views.py
```diff
@@ -61,16 +63,47 @@ class CarListView(LoginRequiredMixin, generic.ListView):
 class CarDetailView(LoginRequiredMixin, generic.DetailView):
     model = Car
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        car = self.object
+        user = self.request.user
+
+        if isinstance(user, Driver):
+            context["is_driver"] = car.drivers.filter(pk=user.id).exists()
```
MNDonut: fetch drivers in advance

taxi/views.py
```diff
@@ -61,16 +63,47 @@ class CarListView(LoginRequiredMixin, generic.ListView):
 class CarDetailView(LoginRequiredMixin, generic.DetailView):
     model = Car
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        car = self.object
+        user = self.request.user
+
+        if isinstance(user, Driver):
```
MNDonut: why do you need this check?
you have LoginRequiredMixin, thus only authenticated users can acces this view, not anonymous
your AUTH_USER_MODEL = Driver, thus request.user is always Driver or I'm missing something?
Fix everywhere

MNDonut (changes requested):

MNDonut (approved):

PR #732, opened 2024-09-18
Solution

Bogdan-Rylov:
Screenshot 2024-09-18 232250
Screenshot 2024-09-18 232349
Screenshot 2024-09-18 232335
Screenshot 2024-09-18 232322
Screenshot 2024-09-18 232304


taxi/forms.py
```diff
@@ -0,0 +1,23 @@
+from django import forms
+from taxi.models import Car, Driver
+from taxi.validators import validate_license_number
```
MNDonut: split different groups of imports

taxi/validators.py
```diff
@@ -0,0 +1,11 @@
+import re
+from django.core.exceptions import ValidationError
```
MNDonut: split different groups of imports

taxi/views.py
```diff
@@ -79,6 +81,24 @@ class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
...
+    return redirect("taxi:car-detail", pk=car.pk)
+
+
+@login_required
+def delete_me_from_car(request, pk):
+    car = get_object_or_404(Car, pk=pk)
+    if request.user in car.drivers.all():
+        car.drivers.remove(request.user)
+
+    return redirect("taxi:car-detail", pk=car.pk)
```
MNDonut: use .exists() so that db uses WHERE and EXISTS at the same time because otherwise you firstly fetch 1 000 000 records and then python loops through all records to check "in" condition

taxi/views.py
```diff
@@ -87,3 +107,23 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    model = Driver
+    fields = "__all__"
+    template_name = "taxi/driver_form.html"
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+class DriverLicenseUpdateView(LoginRequiredMixin, generic.UpdateView):
+    model = Driver
+    form_class = DriverLicenseUpdateForm
+    # fields = ("license_number",)
```
MNDonut: remove this comment

templates/taxi/car_detail.html
```diff
@@ -3,20 +3,36 @@
 {% block content %}
   <h1>
     {{ car.model }}
-    <a href="{% url 'taxi:car-delete' pk=car.id %}" class="btn btn-danger link-to-page">
+  </h1>
+  <p>Manufacturer: ({{ car.manufacturer.name }}, {{ car.manufacturer.country }})</p>
+  <h1>Drivers
+    {% if user in car.drivers.all %}
+      <a href="{% url 'taxi:car-delete-me' car.pk %}" class="btn btn-danger link-to-page">Delete me from this car</a>
+    {% else %}
```

MNDonut (changes requested):

taxi/models.py
```diff
@@ -15,7 +17,11 @@ def __str__(self):
 
 
 class Driver(AbstractUser):
-    license_number = models.CharField(max_length=255, unique=True)
+    license_number = models.CharField(
```
Oleksl888, line 20: Actually the task implies that you make validation for Driver in the form

taxi/validators.py
```diff
@@ -0,0 +1,12 @@
+import re
+
+from django.core.exceptions import ValidationError
+
+
+def validate_license_number(value: str) -> None:
+    pattern = r"^[A-Z]{3}\d{5}$"
+    if not re.match(pattern, value):
```
Oleksl888: you can use built-in RegexValidator instead

taxi/models.py
Oleksl888: Actually the task implies that you make validation for Driver in the form

taxi/views.py
```diff
@@ -79,6 +82,24 @@ class CarDeleteView(LoginRequiredMixin, generic.DeleteView):
     success_url = reverse_lazy("taxi:car-list")
 
 
+@login_required
```
You will need to update car_detail template to use form instead of hyperlink to send request to the CarDetailView to make this work.

templates/taxi/car_detail.html
```diff
@@ -3,20 +3,36 @@
 {% block content %}
   <h1>
     {{ car.model }}
-    <a href="{% url 'taxi:car-delete' pk=car.id %}" class="btn btn-danger link-to-page">
+  </h1>
+  <p>Manufacturer: ({{ car.manufacturer.name }}, {{ car.manufacturer.country }})</p>
+  <h1>Drivers
+    {% if user in car.drivers.all %}
+      <a href="{% url 'taxi:car-delete-me' car.pk %}" class="btn btn-danger link-to-page">Delete me from this car</a>
```
Oleksl888: Add a  form here instead  so it can send a POST request to assign  driver in CarDetailView

Oleksl888 (changes requested):

MNDonut (approved):

PR #772, opened 2024-11-05
Solution

vladyslav-tmf:
Login page:

Screenshot_1

Driver list page:

Screenshot_2

Create driver page:

Screenshot_3

Driver details page;

Screenshot_4

Update driver page:

Screenshot_5

Delete driver page:

Screenshot_6

Car list page:

Screenshot_7

Create car page:

Screenshot_8

Update car page:

Screenshot_9

Delete car page:

Screenshot_10

User assigned to the current car:

Screenshot_11

User not assigned to the current car:

Screenshot_12

Delete manufacturer page:

Screenshot_13

### DIFF
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -1,0 +1,44 @@
+from django import forms
+from django.contrib.auth import get_user_model
+from django.contrib.auth.forms import UserCreationForm
+from django.core.validators import RegexValidator
+
+from taxi.models import Car, Driver
+
+
+class LicenseNumberFormMixin(forms.Form):
+    license_number = forms.CharField(
+        max_length=8,
+        validators=[
+            RegexValidator(
+                regex=r"^[A-Z]{3}[0-9]{5}$",
+                message="The license number must contain 3 capital letters "
+                        "and 5 digits (for example, ABC12345)."
+            )
+        ]
+    )
+
+
+class DriverCreationForm(UserCreationForm, LicenseNumberFormMixin):
+    class Meta(UserCreationForm.Meta):
+        model = Driver
+        fields = UserCreationForm.Meta.fields + (
+            "first_name", "last_name", "license_number"
+        )
+
+
+class DriverLicenseUpdateForm(LicenseNumberFormMixin, forms.ModelForm):
+    class Meta:
+        model = Driver
+        fields = ("license_number",)
+
+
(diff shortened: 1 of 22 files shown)


taxi/forms.py
```diff
@@ -0,0 +1,44 @@
...
+
+class CarForm(forms.ModelForm):
+    drivers = forms.ModelMultipleChoiceField(
+        queryset=get_user_model().objects.all(),
+        widget=forms.CheckboxSelectMultiple,
+    )
+
+    class Meta:
+        model = Car
+        fields = "__all__"
```
AnyoneClown: It's okey to use validation on `Forms` level, but this is best done at the model level. This way, we can make sure that invalid data does not get into our database in any way

[Documentation](https://docs.djangoproject.com/en/5.1/ref/forms/validation/#cleaning-a-specific-field-attribute)

taxi/views.py
```diff
@@ -87,3 +89,32 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    car = get_object_or_404(Car, id=pk)
+    car.drivers.add(request.user)
+    return redirect("taxi:car-detail", pk=pk)
+
+
+@login_required
+def remove_driver(request: HttpRequest, pk: int) -> HttpResponseRedirect:
+    car = get_object_or_404(Car, id=pk)
+    car.drivers.remove(request.user)
+    return redirect("taxi:car-detail", pk=pk)
```
AnyoneClown: We can write this methods in class by rewriting the `post` method

```
class CarRemoveDriverView(LoginRequiredMixin, generic.View):
    def post(self, request, pk):
```

AnyoneClown (changes requested):

taxi/views.py
```diff
@@ -87,3 +89,34 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
 class DriverDetailView(LoginRequiredMixin, generic.DetailView):
     model = Driver
     queryset = Driver.objects.all().prefetch_related("cars__manufacturer")
```
Arsen-hrynevych, line 91: When you use `prefetch_related`, it by default returns all objects that satisfy the rule. Therefore, you don't need to explicitly use `objects.all()`

Arsen-hrynevych (approved):
ok

PR #791, opened 2024-11-23
 Solution-py-taxi-service-user-form-and-validation

skyfoxwork:
Screenshot 2024-11-23 at 16 22 41
Screenshot 2024-11-23 at 16 22 55
Screenshot 2024-11-23 at 16 23 14
Screenshot 2024-11-23 at 16 23 57
Screenshot 2024-11-23 at 16 24 03
Screenshot 2024-11-23 at 16 24 14
Screenshot 2024-11-23 at 16 24 28
Screenshot 2024-11-23 at 16 24 47
Screenshot 2024-11-23 at 16 25 11
Screenshot 2024-11-23 at 16 25 20

### DIFF
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -7,49 +7,20 @@ from django.core.validators import MinLengthValidator, MaxLengthValidator
 from taxi.models import Driver, Car
 
 
-class DriverCreationForm(UserCreationForm):
-    class Meta(UserCreationForm.Meta):
-        model = Driver
-        fields = UserCreationForm.Meta.fields + ("license_number",)
-
+class LicenseNumberValidatorMixin:
     license_number = forms.CharField(
         required=True,
         validators=[MinLengthValidator(8), MaxLengthValidator(8)]
     )
 
     def clean_license_number(self):
+
         license_number = self.cleaned_data["license_number"]
 
-        if (
-                not license_number[:3].isupper()
-                or not license_number[:3].isalpha()
-        ):
+        if len(license_number) != 8:
             raise ValidationError(
-                "First 3 characters of license "
-                "number must be uppercase letters"
+                "License number must be exactly 8 characters long."
             )
-
-        if not license_number[3:].isdigit():
-            raise ValidationError(
-                "Last 5 characters of license number must be digits"
-            )
-
-        return license_number
-
-
-class DriverLicenseUpdateForm(forms.ModelForm):
-
-    license_number = forms.CharField(
-        required=True,
-        validators=[MinLengthValidator(8), MaxLengthValidator(8)]
-    )
-
-    class Meta:
-        model = Driver
-        fields = ("license_number",)
-
-    def clean_license_number(self):
-        license_number = self.cleaned_data["license_number"]
 
         if (
                 not license_number[:3].isupper()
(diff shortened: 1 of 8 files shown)


taxi/forms.py
```diff
@@ -0,0 +1,80 @@
...
+
+class DriverCreationForm(UserCreationForm):
+    class Meta(UserCreationForm.Meta):
+        model = Driver
+        fields = UserCreationForm.Meta.fields + ("license_number",)
+
+    license_number = forms.CharField(
+        required=True,
+        validators=[MinLengthValidator(8), MaxLengthValidator(8)]
+    )
```
vladyslav-tmf, line 14: You declared the `license_number` field and validation 2 times: in the `DriverCreationForm` and `DriverLicenseUpdateForm` classes. This doesn't comply with the **DRY** principle.
Consider declaring this field in a separate class and inheriting it as a **Mixin** class.

taxi/forms.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def clean_license_number(self):
```
vladyslav-tmf, line 16: It's ok, but validating the `license_number` directly in the **model** is a more centralized and efficient approach.
It ensures consistency across all entry points (forms, admin, APIs) and reduces code duplication.
Consider moving your validation logic to the **model** to adhere to the **DRY** principle and improve maintainability.

taxi/urls.py
```diff
@@ -46,6 +51,19 @@
     path(
         "drivers/<int:pk>/", DriverDetailView.as_view(), name="driver-detail"
     ),
+    path("drivers/create/", DriverCreateView.as_view(), name="driver-create"),
+    path(
+        "drivers/<int:pk>/delete",
```
vladyslav-tmf: Ensure all URLs end with `/` to follow Django's best practices and avoid unexpected 404 errors.

taxi/urls.py
```diff
@@ -46,6 +51,19 @@
...
+        DriverDeleteView.as_view(),
+        name="driver-delete"
+    ),
+    path(
+        "drivers/<int:pk>/licens-update",
+        DriverLicenseUpdateView.as_view(),
+        name="driver-update"
```
vladyslav-tmf: Fix the typo in `licens-update` (should be `license-update`) and ensure route names are unique to prevent conflicts (e.g., `driver-update` is already used).

taxi/views.py
```diff
@@ -87,3 +88,38 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+
+class DriverLicenseUpdateView(LoginRequiredMixin, generic.UpdateView):
+    model = Driver
+    form_class = DriverLicenseUpdateForm
+
+
+@login_required
+def assign_me_car(request, pk):
+    if request.method == "POST":
+        driver = Driver.objects.get(id=request.user.id)
```
vladyslav-tmf: Using `Driver.objects.get(id=request.user.id)` can raise a `Driver.DoesNotExist` error if the object is not found.
It's better to use `get_object_or_404`, which automatically handles this case by returning a 404 response.

vladyslav-tmf (changes requested):

taxi/views.py
skyfoxwork: I agree with you but in this case it will work, because I will use this view (assign_me_car, delete_me_car) from car detail page (cars/<int:pk>/) as login user and I will have  car.id and user.id all time.

taxi/forms.py
skyfoxwork: I don't agree with you because, People from other country can have other template for license number (for example it cab be more than 8 characters or it can be all letters or all can be number, letter of upper letter).
if you will be able to change it from admin panel it will better.

PR #800, opened 2024-11-25
solution

alauda6:
Screenshot 2024-11-25 at 14 20 20
Screenshot 2024-11-25 at 14 20 28
Screenshot 2024-11-25 at 14 20 54
Screenshot 2024-11-25 at 14 21 01
Screenshot 2024-11-25 at 14 23 56

### DIFF
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -12,6 +12,12 @@ class DriverCreationForm(UserCreationForm):
         fields = UserCreationForm.Meta.fields + ("license_number",)
 
 
+class DriverLicenseUpdateForm(forms.ModelForm):
+    class Meta:
+        model = Driver
+        fields = ("license_number",)
+
+
 class CarForm(forms.ModelForm):
     drivers = forms.ModelMultipleChoiceField(
         queryset=get_user_model().objects.all(),
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -1,0 +1,22 @@
+from django import forms
+from django.contrib.auth import get_user_model
+
+from taxi.models import Driver, Car
+
+
+class DriverLicenseUpdateForm(forms.ModelForm):
+    class Meta:
+        model = Driver
+        fields = ("license_number",)
+
+
+class CarForm(forms.ModelForm):
+    drivers = forms.ModelMultipleChoiceField(
+        queryset=get_user_model().objects.all(),
+        widget=forms.CheckboxSelectMultiple,
+        required=False,
+    )
+
+    class Meta:
+        model = Car
+        fields = "__all__"
--- a/taxi/forms.py
+++ b/taxi/forms.py
@@ -9,9 +9,7 @@ class DriverCreationForm(UserCreationForm):
 
     class Meta(UserCreationForm.Meta):
         model = Driver
-        fields = UserCreationForm.Meta.fields + (
-           "license_number",
-        )
+        fields = UserCreationForm.Meta.fields + ("license_number",)
 
 
 class CarForm(forms.ModelForm):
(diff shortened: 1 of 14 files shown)


taxi/forms.py
```diff
@@ -0,0 +1,22 @@
...
+class CarForm(forms.ModelForm):
+    drivers = forms.ModelMultipleChoiceField(
+        queryset=get_user_model().objects.all(),
+        widget=forms.CheckboxSelectMultiple,
+        required=False,
+    )
+
+    class Meta:
+        model = Car
+        fields = "__all__"
```
vladyslav-tmf: > On the driver list page create a button that leads to the driver creation page.

You must implement the `DriverCreationForm` using `UserCreationForm` as the task requires.

taxi/urls.py
```diff
@@ -44,7 +47,24 @@
...
+        DriverCreateView.as_view(),
+        name="driver-create"
+    ),
+    path(
+        "drivers/<int:pk>/update/",
+        DriverUpdateView.as_view(),
+        name="driver-update"
+    ),
+    path(
+        "drivers/<int:pk>/delete",
```
vladyslav-tmf:
```suggestion
        "drivers/<int:pk>/delete/",
```
Don't forget to add a `/` at the end of endpoints to avoid unexpectable 404 errors.

taxi/views.py
```diff
@@ -60,12 +61,28 @@ class CarListView(LoginRequiredMixin, generic.ListView):
 
 class CarDetailView(LoginRequiredMixin, generic.DetailView):
```
vladyslav-tmf, line 62: It’s ok, but consider separating the logic of adding and removing drivers into distinct views ( for example, `CarAssignDriverView` and `CarRemoveDriverView`) to follow the **Single Responsibility Principle**, rather than combining it in `CarDetailView`.

taxi/views.py
```diff
@@ -84,6 +101,23 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
     paginate_by = 5
 
 
+class DriverCreateView(LoginRequiredMixin, generic.CreateView):
+    model = Driver
+    fields = "__all__"
+    success_url = reverse_lazy("taxi:driver-list")
```
vladyslav-tmf: You should use `DriverCreationForm`  for driver creation.

taxi/views.py
```diff
@@ -84,6 +101,23 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+
+class DriverUpdateView(LoginRequiredMixin, generic.UpdateView):
+    model = Driver
+    form_class = DriverLicenseUpdateForm
+    success_url = reverse_lazy("taxi:driver-list")
+
+
+class DriverDeleteView(LoginRequiredMixin, generic.DeleteView):
+    model = Driver
+    success_url = reverse_lazy("taxi:car-list")
```
vladyslav-tmf:
```suggestion
    success_url = reverse_lazy("taxi:driver-list")
```
The `success_url` should redirect to the `taxi:driver-list` page, not `taxi:car-list`, because after deleting a driver, the list of drivers should be shown. Fix the typo.

taxi/views.py
```diff
@@ -84,6 +101,23 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+
+
 class DriverDetailView(LoginRequiredMixin, generic.DetailView):
     model = Driver
     queryset = Driver.objects.all().prefetch_related("cars__manufacturer")
```
vladyslav-tmf:
```suggestion
    queryset = Driver.objects.prefetch_related("cars__manufacturer")
```
You can remove `.all()` and just use `Driver.objects.prefetch_related("cars__manufacturer")`.

vladyslav-tmf (changes requested):

taxi/forms.py
alauda6: I missed that info somehow, bad interpretation

PR #808, opened 2024-11-26
Solution

yulia-pl:


dxrrkwm (changes requested):
looks good i guess but attach screenshots

yulia-pl:
image
image
image

taxi/models.py
```diff
@@ -15,7 +17,7 @@ def __str__(self):
 
 
 class Driver(AbstractUser):
-    license_number = models.CharField(max_length=255, unique=True)
+    license_number = models.CharField(max_length=8, unique=True)
```
dxrrkwm, line 20: don't think you need to change it here, you already have validation

taxi/views.py
```diff
@@ -61,10 +68,18 @@ class CarListView(LoginRequiredMixin, generic.ListView):
 class CarDetailView(LoginRequiredMixin, generic.DetailView):
```
dxrrkwm: `   
 model = Driver
    form_class = DriverCreateForm
    success_url = reverse_lazy("taxi:driver-list")
`

you can do it like this so it would be more readable (not sure tho)

dxrrkwm (approved):
looks good overall, check comments :)

PR #879, opened 2025-02-03
empty commit

shretsker:


Chukhno-Maksym (commented):
Implement Create and Delete views for Driver.
You should have forms.py in app "taxi"

Chukhno-Maksym (changes requested):
Implement Create and Delete views for Driver.
You should have forms.py in app "taxi"

PR #910, opened 2025-03-17
Solution

yarzhuravlov:


yarzhuravlov:
Screenshot 2025-03-17 183332
Screenshot 2025-03-17 183336
Screenshot 2025-03-17 183352
Screenshot 2025-03-17 183355
Screenshot 2025-03-17 183241
Screenshot 2025-03-17 183253
Screenshot 2025-03-17 183313

Baranotik15 (approved):
Good Job

taxi/forms.py
```diff
@@ -0,0 +1,48 @@
...
+    )
+
+    class Meta:
+        model = Driver
+        fields = ["license_number"]
+
+
+class CarForm(forms.ModelForm):
+    drivers = forms.ModelMultipleChoiceField(
+        queryset=get_user_model().objects.all(),
```
Boyok1337: It's better to define the user model at the beginning of the file so that you can reuse it right away

taxi/models.py
```diff
@@ -15,7 +16,19 @@ def __str__(self):
...
+        validators=[
+            RegexValidator(
+                regex="^[A-Z]{3}[0-9]{5}$",
+                message="License number should consist "
+                        "only 8 characters where "
+                        "first 3 characters are uppercase letters and "
+                        "last 5 characters are digits."
+            )
+        ]
+    )
```
Boyok1337: Stick to DRY, try to define this validator in the utils file and use it as much as you need and where you need it

Boyok1337 (changes requested):

taxi/models.py
yarzhuravlov: Move license_number_validator to utils and reuse it in models and forms. Thanks for the feedback

taxi/forms.py
yarzhuravlov: Changed it, thanks for the feedback

Boyok1337 (approved):

PR #929, opened 2025-04-03
Implement create, update, delete view for driver + car feature

PrimeGlorious:
image
image
image
image
image


taxi/forms.py
```diff
@@ -0,0 +1,48 @@
...
+from django.core.exceptions import ValidationError
+
+from taxi.models import Driver, Car
+
+
+class DriverCreateForm(UserCreationForm):
+    class Meta(UserCreationForm.Meta):
+        model = Driver
+        fields = (UserCreationForm.Meta.fields
+                  + ("first_name", "last_name", "license_number",))
```
Boyok1337, line 13: License must be vlalidated for the driver creation as well

taxi/views.py
```diff
@@ -1,9 +1,11 @@
 from django.contrib.auth.decorators import login_required
+from django.http import HttpResponseRedirect
 from django.shortcuts import render
 from django.urls import reverse_lazy
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
 
+from .forms import DriverCreateForm, CarForm, DriverLicenseUpdateForm
```
Boyok1337: Use absolute imports instead of relative

Boyok1337 (changes requested):

taxi/forms.py
PrimeGlorious: It is validated
image

oliasymak24 (approved):

PR #930, opened 2025-04-04
Anton Zhukov - Solution

windings-lab:
Cars:
image
image

Drivers:
image
image
image
image
image


windings-lab:
@mateacademy-ai-mentor 
I think you can't see, but I already added a trailing slash to end of the url

PR #946, opened 2025-04-10
Solution

Anollinna:


ntkachenko21 (approved):
good job, but you need to delete old version field with drivers on Car create page

PR #950, opened 2025-04-11
Develop

yarms1:


LLevkovych (changes requested):
You didn't create templates, also didn't create all views

PR #951, opened 2025-04-11
Solution

Kart0l:


taxi/forms.py
```diff
@@ -0,0 +1,58 @@
...
+            raise ValidationError(
+                "License number must be exactly 8 characters."
+            )
+        if not (license_number[:3].isupper() and license_number[:3].isalpha()):
+            raise ValidationError(
+                "First 3 characters must be uppercase letters."
+            )
+        if not license_number[3:].isdigit():
+            raise ValidationError("Last 5 characters must be digits.")
+        return license_number
```
Roksolana-K, line 49: This code is written twice( in class DriverLicenseUpdateForm and DriverCreationForm). You can write it as a function and then just use this function on your classes in order to not repeat yourself.

Roksolana-K (changes requested):
You did a good job! However, I left some comments. Please, check them. 
Also, I can't see the sreenshots of your service, as required on Readme.

PR #952, opened 2025-04-11
Solution

LLevkovych:


taxi/urls.py
```diff
@@ -46,6 +51,18 @@
...
+    path(
+        "drivers/<int:pk>/update/ ",
+        DriverUpdateView.as_view(),
+        name="driver-update"
+    ),
+    path(
+        "drivers/<int:pk>/delete/",
+        DriverDeleteView.as_view(),
+        name="driver-delete"
+    ),
```
Boyok1337: Follow same code style like inline or each item from new line

taxi/views.py
```diff
@@ -1,9 +1,10 @@
 from django.contrib.auth.decorators import login_required
-from django.shortcuts import render
+from django.shortcuts import render, get_object_or_404, redirect
 from django.urls import reverse_lazy
 from django.views import generic
 from django.contrib.auth.mixins import LoginRequiredMixin
 
+from .forms import CarCreationForm, DriverCreationForm, DriverLicenseUpdateForm
```
Boyok1337: Better to use absolute imports instead of relative

taxi/views.py
```diff
@@ -87,3 +88,35 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    model = Driver
+    success_url = reverse_lazy("taxi:driver-list")
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["cancel_url"] = reverse_lazy("taxi:driver-list")
+        return context
+
+
+def assign_to_car(request, pk):
```
Boyok1337: It is better to explicitly specify who will be added to the car

taxi/views.py
```diff
@@ -87,3 +88,35 @@ class DriverListView(LoginRequiredMixin, generic.ListView):
...
+    car = get_object_or_404(Car, pk=pk)
+
+    if request.user in car.drivers.all():
+        car.drivers.remove(request.user)
+    else:
+        car.drivers.add(request.user)
```
Boyok1337: Explicitly add a condition that if the user is not there, then we will add him

Boyok1337 (changes requested):

Boyok1337 (approved):
LGTM!

PR #979, opened 2025-05-05
Solution

olek007:
image
image
image
image
image
image
image
image


olek007:
@mateacademy-ai-mentor 
The condition {% if user in car.drivers.all %} is correct. You can't call function in template
Other issues fixed

PR #985, opened 2025-05-12
Solution

shinodegu:
Снимок экрана (53)
Снимок экрана (54)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
-django==4.1
+django==5.2
 flake8==5.0.4
 flake8-quotes==3.3.1
 flake8-variables-names==0.0.5
 pep8-naming==0.13.2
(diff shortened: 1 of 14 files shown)


shinodegu:
In this task, you will implement a custom form and django built-in forms to create, update or delete content from the site.

Implement:

Create, Delete views for Driver,
On the driver list page create a button that leads to the driver creation page.

Create a driver's license update page. The form - DriverLicenseUpdateForm on this page must check that license:

Consist only of 8 characters
First 3 characters are uppercase letters
Last 5 characters are digits
Also, don't forget to validate the license number on user creation as well.

On the driver detail page add buttons that lead to the driver's license updating page and driver deletion page.

These are assignments to a task. The last task has a condition to add the current user to the car, and only drivers can be registered on the site. So the code passes under the requirements in this case

PR #1047, opened 2025-08-14
Solution

viannik:
image
image
image
image
image
image


Artcrafterrra (approved):
lgmt

viannik:
@mateacademy-ai-mentor The license number validation has already been implemented in the model, so it is already active in all these forms and does not require additional definition.

PR #1057, opened 2025-08-17
Solution

reinvvay:


reinvvay:
{7A517491-091A-4ABB-AEE8-F206B37F4DB1}
{D2064C60-3975-4646-B3DF-D7B959F9A710}
{D098797F-C5A5-4E11-A80B-972F5CB5F581}

taxi/forms.py
```diff
@@ -0,0 +1,57 @@
...
+        if not (
+                license_number[:3].isalpha() and license_number[:3].isupper()
+        ):
+            raise ValidationError(
+                "First 3 characters are not uppercase letters"
+            )
+        if not license_number[3:].isdigit():
+            raise ValidationError("Last 5 characters are not digits")
+
+        return license_number
```
yagginov: It is bad practice to repeat the same code twice.
better rewrite it through a validator. The video was about it. You can also use the already built-in `RegexValidator` from `django.core.validators`. There are examples on the internet on how to work with it ([RegexValidator](https://docs.djangoproject.com/en/5.2/ref/validators/#regexvalidator)) so you can try it, or just take your function as a validator and reuse it

yagginov (changes requested):
only one but quite critical mistake that should be fixed

taxi/forms.py
reinvvay: Thanks for the correction👌

yagginov (approved):
you could just specify this function as a validator for the field, but ok, it's better than it was

reinvvay:
@mateacademy-ai-mentor all files have blank line at the end.

PR #1060, opened 2025-08-17
Solution

Kostenikov:
<details><summary>Screenshots:</summary>
image
image
image
image
image
image
image
image
image
</details>


farzamgt (approved):

Kostenikov:
@mateacademy-ai-mentor First problem solved. For the second - we don't have any other place in templates where user can go to the `driver_confirm_delete.html` page, that means for this case exists only one previous page for `Cancel` button can refer is `taxi:driver-detail` endpoint.

PR #1287, opened 2026-05-20
first code commit

amesemyta1:


amesemyta1:
@mateacademy-ai-mentor 
You're asking me to do something that wasn't in the assignment and has nothing to do with it. I've done everything I was told to do.
In this task, you will implement a custom form and django built-in forms to create, update or delete content from the site.

Implement:

Create, Delete views for Driver,
On the driver list page create a button that leads to the driver creation page.

Create a driver's license update page. The form - DriverLicenseUpdateForm on this page must check that license:

Consist only of 8 characters
First 3 characters are uppercase letters
Last 5 characters are digits
Also, don't forget to validate the license number on user creation as well.

On the driver detail page add buttons that lead to the driver's license updating page and driver deletion page.

PR #1350, opened 2026-07-07
Taxi service validation and user form

Neyrony:
image
image
image
image
image
image
image


Neyrony:
@mateacademy-ai-mentor i disagree with you because model = Driver in admin and views it's best practice than implicitly set model as get_user_model

PR #1377, opened 2026-08-05
Solution

AndriiDraga:


Sn1kls (approved):
Approving. One small cleanup for the future: `LICENSE_PATTERN` is compiled on line 10 in `taxi/forms.py` but never used – the mixin re-defines the pattern locally. Use `LICENSE_PATTERN.match(license_number)` (and drop the local string), or remove the unused constant

PR #1381, opened 2026-08-12
Add driver forms and car assignment

askawa:


askawa:
manufacturer-delete
driver-list
driver-license-update
driver-detail
driver-delete
driver-create
car-detail-delete-me
car-detail-assign-me
car-delete
car-create

